PR #130, opened 2020-06-21
Modifies the third bullet to improve the flow.

SuzeShardlow:
Changes from: To do that, our social network it's as simple as possible
To: To do that, our social network is as simple as possible


AlbertoPdRF (approved):
LGTM, thank you for your first contribution! 🎉

Issue #24, opened 2020-04-30
Add support for social log in/sign up


AlbertoPdRF:
It has to be investigated which is the best way to add this feature.

AlbertoPdRF:
For now I think I'll only add Google and Facebook sign up/log in.

gudeliauskaspam:
What about using twitter too?

AlbertoPdRF:
I just unassigned myself from this one, as I'm kind of blocked at the moment trying to decide on the flow to follow. If anyone wants to discuss about this, let's do it! 😃

PR #293, opened 2020-09-20
Allow users to log in with their Google account

AlbertoPdRF:
I'm opening this as a draft PR because I'm not completely happy yet with the proposed solution. Any code reviews and/or high level comments would be greatly appreciated!

As it is now, this code lets already registered users to log in using their Google account (provided that their email on our database matches their Google's one). This doesn't allow users to register using their Google account though, which is super important. I didn't implement this yet because I'm not sure how should the flow go, so suggestions would be awesome!

This is part of #24.


AlbertoPdRF:
I'm closing this PR, as I have a better flow in mind -- will submit a new one when the code is ready!

PR #309, opened 2020-09-28
Revamped about page with images

### LINKED ISSUE #303: Revamp the `About` component

The `About` component only has text, which is not very visually appealing.

I would like illustrations from [unDraw](https://undraw.co/) to be added to it, so it's more visually appealing.

Leave it as is.

**Additional context**

None.

san-oop:
Added images from UnDraw and the updated few UI for the about page

Fixes #303.


client/src/Components/About/About.js
```diff
@@ -11,7 +11,7 @@ import { useUser } from '../UserProvider/UserProvider';
 
 import Footer from '../Footer/Footer';
 
-import Style from '../Style/Style';
+import Style from './styles';
```
AlbertoPdRF: Instead of creating a new stylesheet, it would be better to keep the already existing one and add there the new necessary styles. Also, if it's just a one-of style, it can be inlined too.

client/src/Components/About/About.js
```diff
@@ -22,45 +22,52 @@ const About = () => {
   const user = useUser();
 
   const formatSection = (section, invertOrder) => {
-    const { title, paragraph, button } = section;
+    const { title, image, paragraph, button } = section;
 
     const getGridItem = (key, content) => {
       return (
-        <Grid item key={key} className={style.center} sm={6} xs={12}>
+        <Grid item key={key} sm={6} xs={12}>
```
AlbertoPdRF: I think that the blocks look better if centered.

client/src/Components/About/About.js
```diff
@@ -22,45 +22,52 @@ const About = () => {
...
     const titleGridItem = getGridItem(
-      title,
-      <Typography variant="h3" color="primary">
-        {title}
-      </Typography>
+      image,
+      <div className={style.imageWrapper}>
+        <img className={style.image} src={image} alt={title} />
+      </div>
     );
```
AlbertoPdRF: Instead of all of this, you could import the SVG directly as a component at the top of the file and then use and style it directly, like:

```javascript
import { ReactComponent as EventsImage } from "../../assets/images/events.svg";

...

<EventsImage className={style.image}/>
```

Also, maybe it looks better to keep the title on this grid item instead of in the text+button one?

client/src/Components/About/About.js
```diff
@@ -22,45 +22,52 @@ const About = () => {
...
-                rel: 'noopener noreferrer',
-              }
-        )}
-      >
-        {button.content}
-      </Button>,
-    ]);
+    const contentGridItem = getGridItem(
+      button.href,
+      <>
```
AlbertoPdRF: For consistency throughout the codebase, please change the first line to `import React, { Fragment } from 'react';` and then use `<Fragment></Fragment>` here.

client/src/Components/About/About.js
```diff
@@ -22,45 +22,52 @@ const About = () => {
...
+        <Typography variant="h4" color="#000000">
```
AlbertoPdRF: The `color` property with a hex code is not valid for the typography component, but please note that is not needed for the title to be rendered black (or white in dark mode).

Any reason why you changed it from `h3` to `h4`?

Also, please note that I suggested moving the title to the grid item with the image, but in either case probably adding the `gutterBottom` property will make everything look a bit cleaner.

client/src/Components/About/About.js
```diff
@@ -22,45 +22,52 @@ const About = () => {
...
+          {title}
+        </Typography>
+        <Typography key={button.content} gutterBottom variant="body1">
```
AlbertoPdRF: The `key` property is no longer needed here.

client/src/Components/About/About.js
```diff
@@ -22,45 +22,52 @@ const About = () => {
...
+          {paragraph}
+        </Typography>
+        <Button
+          key={button.href}
```
AlbertoPdRF: The `key` property is no longer needed here.

client/src/Components/About/About.js
```diff
@@ -22,45 +22,52 @@ const About = () => {
...
+          variant="contained"
+          color="primary"
+          className={style.button}
```
AlbertoPdRF: For consistency with the design of the whole app, this button shouldn't be further styled.

client/src/Components/About/About.js
```diff
@@ -114,13 +125,14 @@ const About = () => {
           formatSection(section, index % 2 !== 0)
         )}
         {!user && (
-          <Grid item className={style.center} xs={12}>
+          <Grid item className={style.footer} xs={12}>
```
AlbertoPdRF: You can leave the `className={style.center}` here and add a `style={{ marginTop: '8px' }}` property to accomplish the same. Please note that we use to use multiples of 8 for spacing.

client/src/Components/About/About.js
```diff
@@ -114,13 +125,14 @@ const About = () => {
...
             <Typography gutterBottom variant="body1">
               {t('about.signUp')}
             </Typography>
             <Button
               variant="contained"
               color="primary"
+              className={style.button}
```
AlbertoPdRF: For consistency with the design of the whole app, this button shouldn't be further styled.

AlbertoPdRF (changes requested):
Thanks so much @san-oop for your first PR! :rocket: 

I left a few comments on the code, and here are some regarding the SVGs:

1. I see that you used the light's theme primary color to fill them, which is awesome! However, please note that when switching to dark mode, the primary color is no longer the same. I think we should have two SVGs for each image, each one filled with each primary color, and render the one that it's appropriate based on the theme the user is using the app in.
2. I quite like all images you chose but the events one, which seems more like people having a meeting. Maybe we can find a better-matching one?
3. All are missing a new line at the end of the file.

And that's all, thanks again for contributing! :tada: Let me know if you need help with anything :slightly_smiling_face:

client/src/Components/About/About.js
san-oop: If we handle themeng as well , we will have to use 8 svg imports here and will have to add more logic right. So thought it would be better to use this way of looping images instead. Please provide your valuable thoughts on this.

client/src/Components/About/About.js
san-oop: h4 looked more cleaner. Please confirm and will update it back to h3 right after your confirmation

client/src/Components/About/About.js
san-oop: its a usual UI pattern to show titles with paragraphs rather than with images right ? Please confirm to update as you suggested( image plus title)

AlbertoPdRF: > If we handle themeng as well , we will have to use 8 svg imports here and will have to add more logic right. So thought it would be better to use this way of looping images instead. Please provide your valuable thoughts on this.

Yes, that's right. I think SVGs can be dynamically modified, so maybe it's worth it to look into that.

> its a usual UI pattern to show titles with paragraphs rather than with images right ? Please confirm to update as you suggested( image plus title)

Ok, let's keep them with the text then!

client/src/Components/About/About.js
AlbertoPdRF: `h4` is alright, but in that case you should put it like `<Typography gutterBottom variant="h4" component="h3">` to keep the HTML's hierarchy.

client/src/Components/About/About.js
```diff
@@ -22,45 +23,48 @@ const About = () => {
   const user = useUser();
 
   const formatSection = (section, invertOrder) => {
-    const { title, paragraph, button } = section;
+    const { title, imageName, paragraph, button } = section;
 
     const getGridItem = (key, content) => {
       return (
-        <Grid item key={key} className={style.center} sm={6} xs={12}>
+        <Grid className={style.center} item key={key} sm={6} xs={12}>
```
AlbertoPdRF:
```suggestion
        <Grid item key={key} className={style.center} sm={6} xs={12}>
```
For consistency.

client/src/Components/About/About.js
```diff
@@ -22,45 +23,48 @@ const About = () => {
...
           {content}
         </Grid>
       );
     };
 
     const titleGridItem = getGridItem(
```
AlbertoPdRF: This should be renamed to `imageGridItem`.

client/src/Components/About/About.js
```diff
@@ -22,45 +23,48 @@ const About = () => {
...
-      title,
-      <Typography variant="h3" color="primary">
-        {title}
-      </Typography>
+      imageName,
+      <SvgIcon name={imageName} style={{ width: '65%', height: '65%' }} />
```
AlbertoPdRF: Wouldn't it be better to call this new component `SvgImage`?

client/src/Components/About/About.js
```diff
@@ -22,45 +23,48 @@ const About = () => {
...
-                target: '_blank',
-                rel: 'noopener noreferrer',
-              }
-        )}
-      >
-        {button.content}
-      </Button>,
-    ]);
+    const contentGridItem = getGridItem(
+      button.href,
```
AlbertoPdRF:
```suggestion
      title,
```

I think that now that we are using the titles here, it's better to use those as the keys.

client/src/Components/About/About.js
```diff
@@ -22,45 +23,48 @@ const About = () => {
...
+      <Fragment>
+        <Typography variant="h4" component="h3" gutterBottom>
```
AlbertoPdRF:
```suggestion
        <Typography gutterBottom variant="h4" component="h3">
```
For consistency.

client/src/Components/About/About.js
```diff
@@ -22,45 +23,48 @@ const About = () => {
   const user = useUser();
 
   const formatSection = (section, invertOrder) => {
-    const { title, paragraph, button } = section;
+    const { title, imageName, paragraph, button } = section;
```
AlbertoPdRF:
```suggestion
    const { imageName, title, paragraph, button } = section;
```
For consistency.

client/src/Components/SvgIcons/SvgIcons.js
```diff
@@ -0,0 +1,34 @@
+import React from 'react';
+import { useTheme } from '@material-ui/core/styles';
+
+import { ReactComponent as EventsImageLight } from '../../assets/images/events_light.svg';
```
AlbertoPdRF: Nitpick: please use `-` instead of `_` for the file names.

client/src/Components/SvgIcons/SvgIcons.js
```diff
@@ -0,0 +1,34 @@
...
+  },
+  dark: {
+    events: EventsImageDark,
+    interests: InterestsImageDark,
+    openSource: OpenSourceImageDark,
+    users: UsersImageDark,
+  },
+};
+
+const SvgIcon = ({ name, ...props }) => {
```
AlbertoPdRF: I normally would write this like:

```javascript
const SvgIcon = (props) => {
  { name, ...rest } = props;
}
```

And then leave a blank line between each line of code, as each of them is doing a pretty different thing.

Besides, as mentioned above, I would rename this component to `SvgImage`, which also implies that any mention to `icon` should be replaced to `image` too.

AlbertoPdRF (changes requested):
Such a great work you are doing here, @san-oop! Thanks so much for it!

I've left a couple of mostly nitpick comments below, but this is almost ready to be merged!

Two more things:
1. Please add a blank line at the end of the SVG files.
2. Please update your original comment to include a line with the `Fixes #303.` text, so the issue will get closed when the PR is merged.

AlbertoPdRF:
As discussed on Slack, I pushed [a commit](https://github.com/Tisn/tisn.app/pull/309/commits/1fc250f2235b18483063ec842218c6665fb8bd13) to demonstrate how to add localized `alt` and `title` attributes to the SVGs.

client/src/Components/About/About.js
```diff
@@ -73,6 +77,7 @@ const About = () => {
     {
       title: t('about.interestsTitle'),
       paragraph: t('about.interestsParagraph'),
+      imageName: 'interests',
```
AlbertoPdRF: I missed this in my earlier review. It's only a nitpick: could you please move all the `imageName` above the corresponding `title`?

AlbertoPdRF (commented):
I've pushed a few minor adjustments to not bother you with another review of nitpicks :slightly_smiling_face: 

I believe this is ready to go, just one last question: you are giving a `marginTop` to the sign up section, wouldn't it be better to extract this into the Style file and apply it to each section? Probably a bit of spacing will make everything look cleaner.

AlbertoPdRF (approved):
LGTM!

Thanks so much for your first contribution @san-oop! 🎉

Issue #279, opened 2020-09-08
Populate the development database with dummy data


AlbertoPdRF:
**Is your feature request related to a problem? Please describe.**

As of now, there's a script that can be used to populate the development database with interests, but it would be better if it also added users, events, etc.

**Describe the solution you'd like**

The script should be modified so it also adds users, events, etc.

**Describe alternatives you've considered**

Add users, events, etc. manually.

**Additional context**

This came up in a Slack conversation with @idiglove :slightly_smiling_face:

Be-Like:
Currently in progress.

AlbertoPdRF:
Assigned! :smiley:

Be-Like:
@AlbertoPdRF Unless you want me working on something else, I can go ahead and start working on populating dummy events. Would you like me to include attendees and posts in the same PR as the events?

AlbertoPdRF:
Working on populating the events sounds like a good next step to me!

And regarding your question, I would say that we should handle a collection population per PR, to try and keep them small and manageable. This way we'll be able to iterate faster ⚡

AlbertoPdRF:
I was thinking that, just as we ask the developer if they want to drop dependent collections when deleting a collection, we could ask them if they want to populate collections when trying to populate dependent collections. As that sentence is a bit convoluted, here are some examples to clarify:
- If a developer tries to populate the users collection without having populated the interests one, present them a prompt asking if they want to populate the interests collection or continue without populating it.
- If a developer tries to populate the events collection without having populated the users collection, present them a prompt asking if they want to populate the users one or abort the events population one, as users are required in this case.

This reminds me that events must have at least one related interest, and I'm not sure now if we had that into account when creating the currently existing scripts, so that's something we need to check! 🕵

PR #312, opened 2020-10-01
Populate the development database with dummy users

Be-Like:
Refer to #279. 

@AlbertoPdRF So far I have only created a script to populate the Users and Tokens associated with those users. Before I went too much further with the other tables, I wanted to get your feedback on the current progress I've made. I added a dev dependency that will auto generate the names for the users.

Please let me know if there are any changes you would like me to make. I plan on creating the other scripts similarly.

I also plan on creating a parent script that will run all the other scripts with one command and to simplify the code.


api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
+#! /usr/bin/env node
+
+// TODO: This will all be in the general js to abide by DRY principal
```
AlbertoPdRF: I wonder if it makes sense to take care of this first, instead of leaving it for later? Maybe that could be done on a separate PR to be merged before this one if you prefer.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+if (!userArgs[0].startsWith('mongodb')) {
+  console.log(
+    'Error: you need to specify a valid MongoDB URL as the first argument'
+  );
+
+  return;
+}
+
+const User = require('../models/User');
+const Interest = require('../models/Interest')
```
AlbertoPdRF: There's a missing semicolon at the end of this line, which is surprising because Prettier is supposed to be ran as a pre-commit hook. Do you know if something weird happened here?

(There are also some other coding style issues that Prettier should take care of, but I won't add comments for them to not clutter the review with them.)

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+const Token = require('../models/Token')
+const crypto = require('crypto');
+
+const async = require('async');
```
AlbertoPdRF: I find it interesting that this isn't conflicting with the `async` keyword 😃

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+    dateOfBirth, country, region,
+    preferredLocale, interests, admin,
+    callback
+) => {
+  const user = new User({
+    name, email, emailConfirmed, emailConfirmedAt,
+    dateOfBirth, country, region,
+    preferredLocale, interests, admin
+  });
+  user.setPassword('Password1');
```
AlbertoPdRF:
```suggestion
  user.setPassword('password');
```
I think it would be better to use `'password'` here, as it's easier to remember.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+  await user.save();
+  const token = createTokenForUser(user);
+  await token.save();
+
+  console.log(`User: ${user}`);
```
AlbertoPdRF:
```suggestion
  console.log(`New user: ${user}`);
```

Nitpick: for consistency with the interests script.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+
+const { uniqueNamesGenerator, names } = require('unique-names-generator');
+const countries = require('country-region-data');
+const locales = ['en', 'es'];
+const users = [];
+
+const createUser = async (
+  name, email, emailConfirmed, emailConfirmedAt,
+    dateOfBirth, country, region,
+    preferredLocale, interests, admin,
```
AlbertoPdRF: I guess we could make this more readable by passing a `userParams` object to the function instead of passing all params individually.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+createTokenForUser = (user) => {
+  return new Token({
+    user,
+    type: 'Email',
+    token: crypto.randomBytes(16).toString('hex')
+  });
+}
+
+const generateUsersArray = async () => {
+  let interests = await Interest.distinct("_id");
```
AlbertoPdRF: I didn't know about the `distinct` method! 😄

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+  const config = {
+    dictionaries: [names]
+  }
+
+  let arr = []
```
AlbertoPdRF:
```suggestion
  const arr = []
```
Nitpick: this should be `const`.

Also, maybe we should name it `usersArray` or something similar, just to be super clear about what's happening here 🙂

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+  await token.save();
+
+  console.log(`User: ${user}`);
+  users.push(user);
+  callback(null, user);
+}
+
+function randomDate(start, end) {
+  return new Date(+start + Math.random() * (end - start));
+}
```
AlbertoPdRF:
```suggestion
const randomDate = (startDate, endDate) =>
  new Date(+startDate + Math.random() * (endDate - startDate));
```
I think it would be cleaner to write this function as a one-line arrow function. Also, maybe name the parameters startDate and andDate to make it clear that they are expected to be dates?

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+    'Error: you need to specify a valid MongoDB URL as the first argument'
+  );
+
+  return;
+}
+
+const User = require('../models/User');
+const Interest = require('../models/Interest')
+const Token = require('../models/Token')
+const crypto = require('crypto');
```
AlbertoPdRF:
```suggestion
```
Actually, you don't need to generate these tokens, as they are only used to verify the users' emails, so for simplicity I would just get rid of everything related to them from this script.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+) => {
+  const user = new User({
+    name, email, emailConfirmed, emailConfirmedAt,
+    dateOfBirth, country, region,
+    preferredLocale, interests, admin
+  });
+  user.setPassword('Password1');
+  await user.save();
+  const token = createTokenForUser(user);
+  await token.save();
```
AlbertoPdRF:
```suggestion
```
This is not needed, see the comment on current lines 16 and 17.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+}
+
+createTokenForUser = (user) => {
+  return new Token({
+    user,
+    type: 'Email',
+    token: crypto.randomBytes(16).toString('hex')
+  });
+}
+
```
AlbertoPdRF:
```suggestion
```
This is not needed, see the comment on current lines 16 and 17.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+const generateUsersArray = async () => {
+  let interests = await Interest.distinct("_id");
+  const config = {
+    dictionaries: [names]
+  }
+
+  let arr = []
+  for (let i = 0; i < 100; i++) {
```
AlbertoPdRF: Maybe the `100` here could be extracted into a variable (with a default value of `100`) that is passed when running the script? Just in case someone wants to generate a different number of users.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+    const name = `${uniqueNamesGenerator(config)} ${uniqueNamesGenerator(config)}`;
```
AlbertoPdRF:
```suggestion
    const name = uniqueNamesGenerator({
      dictionaries: [names, names],
      separator: ' ',
    });
```
I think it would look cleaner to pass the configuration options directly to the function here. I also would change the options as suggested so we can get rid of one of the function calls.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+    const email = `${name.replace(/ /g, '_')}@test-tisn-email.com`.toLowerCase();
```
AlbertoPdRF:
```suggestion
    const email = `${name.replace(/ /g, '_')}@tisn.app`.toLowerCase();
```
You can totally not accept this suggestion 😛

Also, right below this I would define the `emailConfirmed` and `emailConfirmedAt` variables, just so it's totally clear what we are passing to the `createUser` function when calling it below.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+    const country = countries[Math.floor(Math.random() * countries.length)];
+    const region = country.regions[Math.floor(Math.random() * country.regions.length)];
+    const enrolledDate = randomDate(new Date(2019, 01, 01), new Date());
```
AlbertoPdRF: See the suggestion on current line 75 regarding where and how to define this `enrolledDate` (actually `emailConfirmedAt`) variable.

Also, for the `startDate` we could use the date Tisn was launched (2020-05-05) as a sort of easter egg 🤓

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+    const birthDate = randomDate(new Date(1950, 01, 01), new Date(2007, 01, 01));
```
AlbertoPdRF: What about defining `const now = new Date()` above and then using `now.getFullYear() - 13` to define the maximum allowed `endDate`? This now variable could also be used as the `endDate` for the `emailConfirmedAt` variable.

Also, for the `startDate` we could use 1970-01-01 as a sort of easter egg too 🤓

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+    interests = interests.sort(() => 0.5 - Math.random());
+    const selectedInterests = interests.slice(0, Math.floor(Math.random() * 10));
+
```
AlbertoPdRF:
```suggestion
    const selectedInterests = interests
      .sort(() => 0.5 - Math.random())
      .slice(0, Math.floor(Math.random() * interests.length));
```
Two suggestions in one:
1. Consider merging these two lines.
2. Use `interests.length` instead of `10` to set the maximum number of interests a user could be interested in.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+          name,
+          email,
+          true,
+          enrolledDate,
+          birthDate,
+          country.countryShortCode,
+          region.shortCode || 'N/A',
+          locales[Math.floor(Math.random() * 2)],
+          selectedInterests,
+          false,
```
AlbertoPdRF: I think it would be more clear to define the `admin` variable above, as suggested with the `emailConfirmed*` variables on current line 75.

Also, apart from the random users, it would probably make sense to create an admin user for developers to be able to use without having to create one.

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+    arr.push(
+      (seriesCallback) => {
+        createUser(
+          name,
+          email,
+          true,
+          enrolledDate,
+          birthDate,
+          country.countryShortCode,
+          region.shortCode || 'N/A',
```
AlbertoPdRF: This is interesting. I guess if you added the `|| 'N/A'` part is because there are countries without regions, right? If that's the case, we need to open an issue regarding that, as I'm pretty sure it breaks other parts of the app 😖

api/db/populate_users.js
```diff
@@ -0,0 +1,121 @@
...
+const User = require('../models/User');
+const Interest = require('../models/Interest')
+const Token = require('../models/Token')
+const crypto = require('crypto');
+
+const async = require('async');
+
+const mongoose = require('mongoose');
+const mongoDB = userArgs[0];
+mongoose.connect(mongoDB, { useNewUrlParser: true, useUnifiedTopology: true });
```
AlbertoPdRF:
```suggestion
mongoose.connect(mongoDB, {
  useNewUrlParser: true,
  useUnifiedTopology: true,
  useCreateIndex: true,
});
```
Oops! This needs to be updated on the interests script too, as it was done on the `app.js` file as part of 6d2f540a8d7905cf750afba61d50b8f61e561615.

AlbertoPdRF (changes requested):
Wow, thanks for this amazing first contribution @Be-Like! This will for sure be very much appreaciated by any future contributor! 🤩

I've left a bunch of comments for you (sorry if they are too many! 🙏)!

Thanks again for such an amazing

Besides, to follow the convention we are using regarding file names, the new script should be named using `-` instead of `_`, and it would be better to rename the interests script to `populate-interests.js` instead of just `populate.js`.

Thanks again for contributing!  I'll assign #279 to you 🙂

EDIT: oops, it seems I can't assign the issue to you, at least not until you leave a comment on it.

Be-Like:
Thanks for the feedback! I'll take a look at your comments and let you know if I have any questions. I also added a comment to #279 so it can be assigned.

api/db/populate_users.js
Be-Like: I'll take a look at the settings. I don't recall getting any warnings/errors regarding it. I am using the same development settings that I have for work, which may be causing Prettier not to be working correctly.

api/db/populate_users.js
Be-Like: Not a problem, I can break this into another PR if that is what you prefer.

api/db/populate_users.js
Be-Like: Nice catch! I've modified it to avoid any potential issues that could possibly arise.

api/db/populate_users.js
Be-Like: I can do that. Are you planning to make password requirements anytime in the future (ie. must contain 1 letter, 1 number, 1 special character)? If so, I would recommend we set the password to be in line with those potential requirements now rather than having a potential issue in the future.

api/db/populate_users.js
Be-Like: Works for me. This has been resolved.

api/db/populate_users.js
Be-Like: Yeah, this was a new one for me as well. It returns an array of of all the specified fields without their key value.

api/db/populate_users.js
Be-Like: Works for me. 👍

api/db/populate_users.js
AlbertoPdRF: Feel free to go either way! :slightly_smiling_face:

api/db/populate_users.js
AlbertoPdRF: I have no plans to do that as of now, so let's keep it simple for the time being.

api/db/populate_users.js
AlbertoPdRF: I forgot one thing here: maybe it would be better to name the function `getRandomDate`, just to follow the conventions.

AlbertoPdRF:
Wonderful! :100:

By the way, let's not forget to keep [the API's README](https://github.com/Tisn/tisn.app/blob/master/api/README.md) updated with the changes that you are doing. A good addition after all of the work is done would be to add an npm script named `populate-db` or something similar that takes care of populating the database running the corresponding scripts. This way, the user doesn't need to run the command listed on the API's README, and we could also move that new instruction to the main one!

api/db/populate_users.js
Be-Like: Very good suggestions!

api/db/populate_users.js
Be-Like: How would you like that to be handled? I can make it a flag to be passed in the command line. Or I can have 1 (or any number) admin created each time.

api/db/populate_users.js
Be-Like: That is correct, there are some countries that do not have any regions. 

I was going to bring that one up as a bug along with another regarding emails (see #313).

api/db/populate_users.js
AlbertoPdRF: I think I would just create one admin user and document its existence so developers know of it and how to use it. Also, maybe it's a good idea to use some very clear values for it: `Admin` as its name, `admin@tisn.app` as its email, Tisn's logo as its avatar, etc. The last example just reminded me of something: should we try and use random images from somewhere for the dummy users' avatars, or is that an overkill?

api/db/populate_users.js
AlbertoPdRF: Sounds good! Let's take care of that on its own issue then 😊

api/db/populate_users.js
Be-Like: I don't think it is overkill, though unless it is something super simple it might be worth opening a feature issue for it. 

As for logging to the user, it might be worth looking at a better way of managing and displaying important information to the user. Maybe by default we don't display anything but pertinent information (number of records created, admin information, etc.). And then we can add an optional command line flag for displaying all logs.

Be-Like:
So we can get this PR done more quickly (especially for any others that may be joining for hacktoberfest) and to avoid adding too much complexity to this PR, I think we should open some other feature tickets to expand on this one. 

* Adding command line flags and options 
* Simplifying what is logged
* Image generator

AlbertoPdRF:
That sounds good to me @Be-Like!

api/db/populate_users.js
AlbertoPdRF: For the avatars, we can hopefully use some service like [Adorable Avatars](http://avatars.adorable.io/) quite easily! That's the first thing I found when googling, I'm just dropping the link here.

api/db/populate_users.js
Be-Like: I just pushed up all the updates I've made so far for populating the users. When you get some free time, do you mind checking if the Prettier linting looks better?

api/db/populate_users.js
AlbertoPdRF: We have #314 for this now, so I'm resolving this conversation.

api/db/populate_users.js
AlbertoPdRF: Looks much better, although there are still some things which aren't as they should, for instance the `"`s should be `'`s, and there are still missing semicolons 😕

api/db/populate_users.js
AlbertoPdRF: Are you opening this issue, or should I? 🙂

AlbertoPdRF:
Ok, this looks quite good to me! Let's resolve the remaining Prettier issues and get it merged 🚀 Then we can go with all the follow up issues 😄

api/db/populate-users.js
```diff
@@ -0,0 +1,133 @@
...
+}
+
+const User = require("../models/User");
+const Interest = require("../models/Interest");
+
+const asynchronous = require("async");
+
+const mongoose = require("mongoose");
+const mongoDB = userArgs[0];
+const numberOfRecords = userArgs[1] || 100;
```
AlbertoPdRF: Let's better move this down, after all the DB stuff.

api/db/populate-users.js
```diff
@@ -0,0 +1,133 @@
...
+  createUser(
+    {
+      name: "Admin",
+      email: `admin@tisn.app`,
+      emailConfirmed: true,
+      emailConfirmedAt: new Date(),
+      country: "US",
+      region: "FL",
+      preferredLocale: "en",
+      dateOfBirth: new Date(),
```
AlbertoPdRF: This won't be a valid `dateOfBirth`, as the user's age needs to be more than 13.

AlbertoPdRF (changes requested):
Just two small comments, sorry I didn't see this earlier!

api/db/populate_users.js
Be-Like: Just pushed up the reformatted code. Hopefully I've gotten the linter sorted out as it should.

api/db/populate_users.js
Be-Like: If you don't mind doing that. You may have better insight as to where else in the code it may be causing issues.

api/db/populate_users.js
AlbertoPdRF: It only gives me two changes now: the semicolons on current lines 12 and 116 shouldn't be there.

By the way, do you know any way I can make it easier for other contributors to not face these problems? I can see how this can be frustrating for people wanting to contribute 😞

api/db/populate_users.js
AlbertoPdRF: Ok, will do! 🙂

api/db/populate_users.js
Be-Like: Not sure exactly. When I run the commit, I get this error: 
Some of your tasks use `git add` command. Please remove it from the config since all modifications made by tasks will be automatically added to the git commit index.

AlbertoPdRF (approved):
LGTM! Thanks so much for this awesome first contribution @Be-Like! 📈

api/db/populate_users.js
Be-Like: Awesome I will make note of it and look into it.

api/db/populate_users.js
AlbertoPdRF: I'll fix that now, we'll see if it helps! 🙂

api/db/populate_users.js
AlbertoPdRF: Oh, it seems this won't be an issue in the end, as countries with no regions seem to have this structure:

```javascript
{
  countryName: countryName
  countryShortCode: countryShortCode
  regions: [{
    name: countryName
    shortCode: countryShortCode
  }]
}
```

So I think we can just get rid of the ` || 'N/A'` part! 🙂

Be-Like: Sorry, I missed this. That may be the case, but when I left the field null it would fail to create the document. It is likely because the model is requiring the region to have something. Would you like me to remove that `required` flag instead?

```
const UserSchema = new Schema(
  {
    name: { type: String, required: true },
    email: { type: String, required: true },
    emailConfirmed: { type: Boolean, required: true, default: false },
    emailConfirmedAt: { type: Date },
    salt: { type: String, required: true },
    hash: { type: String, required: true },
    dateOfBirth: { type: Date, required: true },
    country: { type: String, required: true },
    region: { type: String, required: true },
    preferredLocale: { type: String, required: true, enum: ['en', 'es'] },
    avatar: { type: String },
    interests: [{ type: Schema.Types.ObjectId, ref: 'Interest' }],
    admin: { type: Boolean, required: true, default: false },
  },
  { timestamps: true }
);

AlbertoPdRF: Ah, I see the problem now -- the problem is not countries without regions, but regions without short codes instead. How does using `region: region.shortCode || region.name` sound?

Be-Like: That works for me. Good eye! I'll make that change now.

AlbertoPdRF: Perfect! I'll open an issue regarding this now, as it's indeed a problem in other parts of the app too.

Issue #314, opened 2020-10-03
Create Master Database Populate Command


Be-Like:
**Is your feature request related to a problem? Please describe.**
Make a master file that will handle the database connection and population of all and/or specified tables.

**Describe alternatives you've considered**
Having separate scripts to run for each individual table.

**Additional context**
@AlbertoPdRF This is with regards to our conversation on the current pull request for the users db. It will be simpler to create a separate PR for this and then individual PR's for each table script.

AlbertoPdRF:
Thanks for the issue @Be-Like! 🙂

Just to keep everyone in sync, this is related to #279.

Will you want to handle this one @Be-Like? 😄

Be-Like:
If that is alright. I already did some research into it and have a good idea of the direction I will take.

AlbertoPdRF:
Wonderful! I just assigned it to you! 🚀

Be-Like:
@AlbertoPdRF, do you mind adding the "hacktoberfest" topic to the about portion of your project? That is, if you are open to having people participate during hacktoberfest. They changed their rules as of today that requires project to "opt-in" to hacktoberfest by including that topic so as to prevent projects from getting spammed.

AlbertoPdRF:
I just did @Be-Like! 😃

PR #316, opened 2020-10-03
Update API's README to acknowledge `populate-users` script

Be-Like:
To reflect the changes related to #279.

### DIFF
--- a/api/README.md
+++ b/api/README.md
@@ -2,4 +2,4 @@
 
 The back end is currently hosted on [Heroku](https://heroku.com/).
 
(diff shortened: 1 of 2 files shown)


AlbertoPdRF:
Thanks for the PR @Be-Like! 💯

Should we wait to make the changes to the READMEs in the PR for #314? As we will be constantly adding more scripts, this way we don't have to keep updating them.

Be-Like:
That works for me. You can go ahead and reject this PR.

AlbertoPdRF:
Ok, let's forget about my previous comment. The main README will only need to be updated when #314 is taken care of, we can keep the back end one in sync while we keep adding scripts.

We can batch the README update with [this change](https://github.com/Tisn/tisn.app/pull/312#discussion_r499178240) and just merge it!

AlbertoPdRF (approved):
LGTM!

Merging now :man_technologist:

PR #317, opened 2020-10-04
Dynamic colors for svg based on the active theme

### LINKED ISSUE #315: Change SVGs main color dynamically

#309 introduced some SVGs that are used on the `About` component. To match the theme the user is using the app in, two SVGs of each type were added with different main colors.

It would be better to just have one SVG of each type, and change its main color dynamically to match the theme the user is using the app in.

Leave the two SVGs of each type.

**Additional context**

None.

san-oop:
Repeated svgs are removed and updated "Fill" attribute of svgs. Now the color of svgs will be updated based on the active theme

Fixes #315


AlbertoPdRF (approved):
Thank you for this PR @san-oop! 🤩

LGTM, I'll merge it right now!

PR #320, opened 2020-10-05
Add database population master script

### LINKED ISSUE #314: Create Master Database Populate Command

Be-Like:
Fixes #314.

This is still a work in progress, however I wanted you to be able to see the progress I have made and give you an opportunity to make any suggestions or corrections. 

Also, if the user runs the command to populate all tables, should we delete and replace the Interests table if it already exists? If we don't, then the interests table would contain duplicates.

The new command for populating all the databases would be:
`node db/populate`

To populate a specific table:
`node db/populate -t <table-name>` or `node db/populate --table <table-name>`

Other flags:
`-v` or `--verbose` to display all logs.
`--number` to specify the number of records to create. (Default: 100; currently only used for the users table)

I'll update the readme with the similar information before we merge this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,11 +62,12 @@ Before starting, copy the `.env-template` file on the `api` folder to a new `.en
 - `SENDGRID_API_KEY`: you can sign up on [SendGrid](https://sendgrid.com/) using the free tier, however this key is only necessary if you want to send emails, which by default are logged to the console in development mode
 - `BASE_CLIENT_URL`: should be `http://localhost:3000` unless you change the port on which the client runs
 
-Once that's done, you can install all the dependencies and launch the back end executing the following commands from the repository's root folder:
+Once that's done, you can install all the dependencies, populate the database, and launch the back end executing the following commands from the repository's root folder:
 
 ```bash
 cd api
 npm install
+npm run db:populate
 npm start
 ```
 
--- a/api/README.md
+++ b/api/README.md
@@ -2,4 +2,41 @@
 
 The back end is currently hosted on [Heroku](https://heroku.com/).
 
-The production database is currently hosted on [MongoDB Atlas](https://cloud.mongodb.com/). There's a script that can be used to populate the database with interests and users. To run them, use the commands `node db/populate-interests 'MONGODB_URL'` and `node db/populate-users 'MONGODB_URL'` respectively.
+The production database is currently hosted on [MongoDB Atlas](https://cloud.mongodb.com/). There's a script that can be used to populate the database with interests and users. To run the script and to populate all collections, use the command `npm run db:populate`.
+
+## Advanced Options
+
+The following flags can be used when populating the database:
+
+- `-c` or `--collection`, specify the collection you wish to populate.
+- `-m` or `--multiplier`, multiply the default number of users generated.
+- `-r` or `--random-location`, randomizes the country and region each user is located.
+- `-v` or `--verbose`, logs each document generated.
+
+Example:
+
+`npm run db:populate -- -c users -m 5 --verbose`
+
+This example will populate the users collection with 500 users and will log to the console each user created.
+
+When the users collection is first populated it also generates an admin user with the following information:
+
+- Name: Admin
+- Email: admin@tisn.app
+- Password: password
+
+Each user's password generated from the script is `password`.
+
+**Note:** Default number of users generated is 100 and the default location each user is from is Madrid, Spain.
+
+## Deleting Collections
+
+To drop the database, run `npm run db:drop`. This will drop all collections in the database.
+
+The following flags can be used when deleting the database:
+
+- `-c` or `--collection`, specify the collection you wish to delete.
+
+Example:
+
+`npm run db:drop -- -c users`
--- a/api/README.md
+++ b/api/README.md
@@ -2,41 +2,33 @@
 
 The back end is currently hosted on [Heroku](https://heroku.com/).
 
-The production database is currently hosted on [MongoDB Atlas](https://cloud.mongodb.com/). There's a script that can be used to populate the database with interests and users. To run the script and to populate all collections, use the command `npm run db:populate`.
+## Database
 
-## Advanced Options
+The applications uses [MongoDB](https://www.mongodb.com/) and [mongoose](https://mongoosejs.com/). The production database is currently hosted on [MongoDB Atlas](https://cloud.mongodb.com/).
 
-The following flags can be used when populating the database:
+### Populate collections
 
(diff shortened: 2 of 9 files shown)


Be-Like:
@AlbertoPdRF Do you want me to update the readme in this PR or do all the readme updates in the #316? Also, do you want me to create the scripts for the other tables in this PR? Or open new issues for each table script?

AlbertoPdRF:
Sorry, I can't provide a full review now, but I'll try to answer your high level comments so you can keep working on this!

But before anything, thanks for the good work! :tada:

> Also, if the user runs the command to populate all tables, should we delete and replace the Interests table if it already exists? If we don't, then the interests table would contain duplicates.

That's a good question! Maybe we should prompt the developer and let them decide between aborting or going ahead dropping the collection of interests and creating a new one? Also, we'll have a similar problem with the admin user, as trying to create another one with the same data will throw an error.

> The new command for populating all the databases would be: `node db/populate`

This is very nice :100: I think it would be awesome to also add an npm script which just runs that command. And maybe we should add a script to drop the database?

> To populate a specific table: `node db/populate -t <table-name>` or `node db/populate --table <table-name>`

And this is also awesome! :top: However, as MongoDB doesn't have `tables` but `collections`, I would rename the argument accordingly.

> Other flags:
> -v or --verbose to display all logs.
> --number to specify the number of records to create. (Default: 100; currently only used for the users table)

This sounds good to me. I'm curious: why is the number treated as a string? And a suggestion: as the number will be used for other collections too, maybe it would be better to use it as a multiplier? This way, we can define a proportion (e.g. 100 users -> 500 events) and then the developer can decide how many the actually want with the multiplier.

> @AlbertoPdRF Do you want me to update the readme in this PR or do all the readme updates in the #316? Also, do you want me to create the scripts for the other tables in this PR? Or open new issues for each table script?

I think we can ship #316 quickly with the two things we have discussed there, and we can then handle the other README updates here. Regarding the creation of the other scripts, let's handle those in separate PRs, as this one will become huge if we do that here! I don't think we need more issues though, as #279 should already be enough :slightly_smiling_face:

AlbertoPdRF:
Another high level comment: as we are creating users in random countries and regions, testing some of the functionality of the app might be difficult, as having more than one user in the same country and region by random chance is unlikely. Maybe we should provide a default country and region for all users (e.g. Madrid, Spain), but allow the developer to override that default or let the users be created in random countries and regions. Please note that this will also apply to events.

Be-Like:
This is now ready for review. 

I changed the `-t` and `--table` flags to `-c` and `--collection`. 

I've made all the users generated to be within the same country and origin. This can be overridden with the `-r` or `--random-country` flag.

I've added y/N console prompts for the following:
1. If the interests collection already exists, it will prompt the user to drop the current collection to avoid duplication of interests. However, user has the ability to allow for the duplicate interest.
2. When creating the users collection, if the interests collection does not exist, then the user will receive a prompt stating that the users generated will not have any interests and ask if they want to continue anyway.

AlbertoPdRF (changes requested):
This is looking very nice @Be-Like! Thanks for all your work on this titanic task :muscle:

As providing a code review within such large diff was going to be difficult, I made some adjustments and commited them to a branck on my fork, see: https://github.com/AlbertoPdRF/tisn.app/commit/80e779863bd88745c26e9f0f3a27e8711e6543fa

Apart from some nitpicks, the big things I changed are:
- Ran `npm install minimist --save-dev` so it's clear that it's a dependency. Please note that this also reordered the `prompts` dependency
- Got rid of the `admin` argument. In the end, developers will need one, and after one is created the argument is not relevant anymore because none more will be created, so I think it doesn't have that much use after all
- Changed the `number` argument to be `multiplier`, as this will be used for events too and allows us to define a proportion instead of just using the same number for both users and events
- Added the `db:populate` npm script so developers can just run `npm run db:populate` to populate the development database. I also added this to the main README
- As it doesn't make much sense to have duplicated interests, I changed the already existing interests collection prompt to be:
  - Yes: drop the existing collection and create a new one, letting the developer know that this can potentially break the app (this will be the case if there are users (and later, events) with those interests)
  - No: do not drop the existing collection nor create a new one

If you agree with all the changes, you can just cherry-pick that commit and add it to your branch, or I can do that myself if you prefer. Also, we can definitely discuss about any of the changes before proceeding!

Things left to do:
- Document the new script together with all its possible options on the API's README
- Decide if we want to add a command to allow the developer to drop the database easily (which could be useful in the case the developer breaks the app :stuck_out_tongue:)
- There's still some code duplication on the `populate.js` file, but I think we can just decide to live with that

And I think that's all, great work @Be-Like!! :star_struck:

Be-Like:
All that sounds great! The only thing I would suggest is maybe having both a number flag and a multiplier flag just in case the user wants to generate more or less users. Otherwise, if you want to generate a 1000 users you would have to run the script 10 times.

Be-Like:
Also, I had originally put those dependencies under the devDependencies because I figure these scripts will only ever be used by developers and we won't need them in production. I don't want to bog down the production build with unnecessary dependencies if we don't have to.

AlbertoPdRF:
> All that sounds great! The only thing I would suggest is maybe having both a number flag and a multiplier flag just in case the user wants to generate more or less users. Otherwise, if you want to generate a 1000 users you would have to run the script 10 times.

In that case, you just run the script with a multiplier of 10 and 1000 users will be generated :slightly_smiling_face:

> Also, I had originally put those dependencies under the devDependencies because I figure these scripts will only ever be used by developers and we won't need them in production. I don't want to bog down the production build with unnecessary dependencies if we don't have to.

Right! That's the way to go, please note that this is still the case!

Be-Like:
Gotcha, thanks for the clarification! I'll go ahead and cherry pick, update the readme, and add the DB drop.

AlbertoPdRF:
That sounds good to me! :smile: Let me know if you need anything! And by the way, did you know about our Slack workspace? There's an invitation link on the README in case you want to join! :slightly_smiling_face:

AlbertoPdRF (approved):
This is some solid work @Be-Like! 💯

I added a commit with some nitpicks to not bother you with more code reviews, so this is ready to be shipped! 🚢

There's some code duplication in both the `populated.js` and `drop.js` files, but we can take care of that when adding the remaining individual scripts.

Thanks again for your work! 🎉

PR #327, opened 2020-10-06
Take into account that JS enumerates months from 0 to 11

AlbertoPdRF:
While reviewing #312 I totally forgot that JavaScript enumerates months from 0 to 11 instead of from 1 to 12. This PR takes that into account!

CC: @Be-Like 👀


Be-Like:
You're right, those should just be 0 to 11 rather than 01. 👍

PR #329, opened 2020-10-07
Add script to populate (and drop) the events collection

Be-Like:
Refer #279. Populate and drop events.

Things to note:
1. When dropping the events collection, I also have it dropping the attendants collection as it is completely dependent on the event.
2. When populating the events, I worked under the assumption that the start dates would range between now (`new Date()`) and 5 years from now (`futureDate`). End date ranges between start date and `futureDate`. 
3. I refactored a lot of code that is and will be reusable between all the population scripts and put them in a `helper-functions.js` file. 
4. I set the maximum attendantsLimit to the total number of users at the time of population, though I know it is arbitrary. I did this so we can more easily test and create functionality around having full events. If we were to set it to be a random number between 1 and 1 million, it would be difficult to find events to test such functionality.

### DIFF
--- a/api/db/drop.js
+++ b/api/db/drop.js
@@ -5,6 +5,7 @@ const { connectDb, closeDb } = require('./connection');
 
 const Interest = require('../models/Interest');
 const User = require('../models/User');
+const Event = require('../models/Event');
 
 const userArgs = minimist(process.argv.slice(2), {
   string: 'collection',
@@ -33,10 +34,21 @@ const dropUsers = async () => {
   }
 };
 
+const dropEvents = async () => {
+  console.log('\n', '\x1b[0m', 'Dropping events collection...');
+  if ((await Event.countDocuments()) !== 0) {
+    await Event.collection.drop();
+    console.log('\x1b[31m', 'Dropped events collection');
+  } else {
+    console.log('\x1b[33m', 'events collection is already empty');
+  }
+};
+
 const dropCollections = async () => {
   connectDb();
   await dropInterests();
   await dropUsers();
+  await dropEvents();
   closeDb();
 };
 
@@ -49,9 +61,12 @@ const dropCollection = async () => {
     case 'users':
       await dropUsers();
       break;
+    case 'events':
+      await dropEvents();
+      break;
     default:
       console.log(
-        `Unknown collection '${userArgs.c}', possible options are: [interests, users]`
+        `Unknown collection '${userArgs.c}', possible options are: [interests, users, events]`
       );
       break;
   }
--- a/api/db/drop.js
+++ b/api/db/drop.js
@@ -23,22 +23,14 @@ const userArgs = minimist(process.argv.slice(2), {
 
 const dropInterests = async () => {
   console.log('\n', '\x1b[0m', 'Dropping interests collection...');
-  if (await getInterestsCount() !== 0) {
+  if ((await getInterestsCount()) !== 0) {
     await Interest.collection.drop();
     console.log('\x1b[31m', 'Dropped interests collection');
   } else {
     console.log('\x1b[33m', 'Interests collection is already empty');
   }
 
-  if (await getEventsCount() !== 0) {
-    const drop = await createPrompt(
-      'The events collection is dependent on the interests collection. Drop events collection?'
-    );
-
-    if (drop) await dropEvents(drop);
-  }
-
-  if (await getUsersCount() !== 0) {
+  if ((await getUsersCount()) !== 0) {
     const drop = await createPrompt(
       'The users collection is dependent on the users collection. Drop users collection?'
     );
@@ -49,7 +41,7 @@ const dropInterests = async () => {
 
 const dropUsers = async (confirmed = false) => {
   console.log('\n', '\x1b[0m', 'Dropping users collection...');
-  if (confirmed || await getUsersCount() !== 0) {
+  if (confirmed || (await getUsersCount()) !== 0) {
     await User.collection.drop();
     console.log('\x1b[31m', 'Dropped users collection');
   } else {
@@ -57,10 +49,10 @@ const dropUsers = async (confirmed = false) => {
     return;
   }
 
-  if (await getEventsCount() !== 0) {
+  if ((await getEventsCount()) !== 0) {
     const drop = await createPrompt(
       'The events collection is dependent on the users collection. Drop events collection?'
-    )
+    );
 
     if (drop) await dropEvents(drop);
   }
@@ -68,7 +60,7 @@ const dropUsers = async (confirmed = false) => {
 
 const dropEvents = async (confirmed = false) => {
   console.log('\n', '\x1b[0m', 'Dropping events collection...');
-  if (confirmed || await getEventsCount() !== 0) {
+  if (confirmed || (await getEventsCount()) !== 0) {
     await Event.collection.drop();
     console.log('\x1b[31m', 'Dropped events collection');
   } else {
@@ -76,7 +68,7 @@ const dropEvents = async (confirmed = false) => {
     return;
   }
 
-  if (await getAttendantsCount() !== 0) {
+  if ((await getAttendantsCount()) !== 0) {
     const drop = await createPrompt(
       'The attendants collection is dependent on the events collection. Drop attendants collection?'
     );
@@ -87,13 +79,13 @@ const dropEvents = async (confirmed = false) => {
 
 const dropAttendants = async (confirmed = false) => {
   console.log('\n', '\x1b[0m', 'Dropping attendants collection...');
-  if (confirmed || await getAttendantsCount() !== 0) {
+  if (confirmed || (await getAttendantsCount()) !== 0) {
     await Attendant.collection.drop();
     console.log('\x1b[31m', 'Dropped attendants collection');
   } else {
     console.log('\x1b[33m', 'Attendants collection is already empty');
   }
-}
+};
 
 const dropCollections = async () => {
   connectDb();
--- a/api/db/drop.js
+++ b/api/db/drop.js
@@ -2,17 +2,17 @@
 
 const minimist = require('minimist');
 const { connectDb, closeDb } = require('./connection');
+const { createPrompt } = require('./utils');
 
 const Interest = require('../models/Interest');
 const User = require('../models/User');
 const Event = require('../models/Event');
 const Attendant = require('../models/Attendant');
-const { createPrompt } = require('./utils');
 
-const interestCount = async () => await Interest.countDocuments();
-const userCount = async () => await User.countDocuments();
-const eventCount = async () => await Event.countDocuments();
-const attendantCount = async () => await Attendant.countDocuments();
+const getInterestsCount = async () => await Interest.countDocuments();
+const getUsersCount = async () => await User.countDocuments();
+const getEventsCount = async () => await Event.countDocuments();
+const getAttendantsCount = async () => await Attendant.countDocuments();
 
 const userArgs = minimist(process.argv.slice(2), {
   string: 'collection',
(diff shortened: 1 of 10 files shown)


api/db/helper-functions.js
```diff
@@ -0,0 +1,39 @@
...
+    initial: true,
+  });
+
+  return answer.value;
+};
+
+const getRandomDate = (startDate, endDate) =>
+  new Date(+startDate + Math.random() * (endDate - startDate));
+
+const getRandomCountry = (randomLocation) =>
```
AlbertoPdRF: As the returned country won't always be random, I would rename the function to `getCountry`. The parameter could also be named just `random`, but I don't feel strongly about that.

api/db/helper-functions.js
```diff
@@ -0,0 +1,39 @@
...
+
+const getRandomDate = (startDate, endDate) =>
+  new Date(+startDate + Math.random() * (endDate - startDate));
+
+const getRandomCountry = (randomLocation) => 
+  randomLocation 
+    ? countries[Math.floor(Math.random() * countries.length)]
+    : { countryShortCode: 'ES' };
+
+const getRandomRegion = (randomLocation, country) =>
```
AlbertoPdRF: Same as above but with `randomRegion` in this case.

api/db/populate-events.js
```diff
@@ -0,0 +1,97 @@
...
+  getRandomDate,
+  getRandomCountry,
+  getRandomRegion ,
+  getRandomInterests
+} = require('./helper-functions');
+
+const Interest = require('../models/Interest');
+const User = require('../models/User');
+const Event = require('../models/Event');
+const { uniqueNamesGenerator, adjectives, animals, starWars, colors } = require('unique-names-generator');
```
AlbertoPdRF: I would move this to line 2.

api/db/populate-events.js
```diff
@@ -0,0 +1,97 @@
...
+
+    const eventParams = {
+      name,
+      description,
+      startDate,
+      endDate,
+      country: country.countryShortCode,
+      region: region.shortCode,
+      user: usersList[Math.floor(Math.random() * usersList.length)],
+      interests: getRandomInterests(interestsList),
```
AlbertoPdRF: For consistency, let's use `relatedInterests` here. Please note that this will also require a change on current line 24.

api/db/populate-users.js
```diff
@@ -1,10 +1,9 @@
-const prompts = require('prompts');
+const { createPrompt, getRandomDate, getRandomCountry, getRandomRegion, getRandomInterests } = require('./helper-functions');
 
 const User = require('../models/User');
 const Interest = require('../models/Interest');
 
 const { uniqueNamesGenerator, names } = require('unique-names-generator');
```
AlbertoPdRF: Let's move this to line 2 (and leave a blank line before the models' imports).

api/package.json
```diff
@@ -48,7 +48,8 @@
     "nodemon": "^2.0.4",
     "prettier": "2.1.2",
     "prompts": "^2.3.2",
-    "unique-names-generator": "^4.3.1"
+    "unique-names-generator": "^4.3.1",
+    "txtgen": "^2.2.7"
```
AlbertoPdRF: This should go above the `unique-names-generator` dependency, as this is automatically ordered alphabetically.

api/db/drop.js
```diff
@@ -33,10 +35,28 @@ const dropUsers = async () => {
   }
 };
 
+const dropEvents = async () => {
```
AlbertoPdRF: I would keep this as a function that only drops the events collection. I'll explain why on a top-level comment.

api/db/drop.js
```diff
@@ -33,10 +35,28 @@ const dropUsers = async () => {
...
+  console.log('\n', '\x1b[0m', 'Dropping events and attendants collection...');
+  if ((await Event.countDocuments()) !== 0) {
+    await Event.collection.drop();
+    console.log('\x1b[31m', 'Dropped events collection');
+  } else {
+    console.log('\x1b[33m', 'events collection is already empty');
```
AlbertoPdRF: The sentence should start with a capital letter.

api/db/populate-events.js
```diff
@@ -0,0 +1,97 @@
...
+      coverPhoto: "",
+      attendantsLimit: Math.floor(Math.random() * usersList.length -1) + 1
```
AlbertoPdRF: If we leave this as is we could have invalid events, as the `attendantsLimit` must be at least 2. Unless I got something wrong, this should be written as: `Math.floor(Math.random() * (usersList.length - 2)) + 2`.

api/db/populate-events.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }
+
+  displayLogs = verbose;
+  const eventsArray = [];
+  const now = new Date();
+  const futureDate = new Date().setFullYear(now.getFullYear() + 5);
+
+  for (let i = 0; i < 500 * multiplier; i++) {
+    const name = uniqueNamesGenerator({
```
AlbertoPdRF, line 71: I would capitalize the first letter of the name.

api/db/populate-events.js
```diff
@@ -0,0 +1,97 @@
...
+      dictionaries: [adjectives, animals, colors, starWars],
+      length: 3,
+      separator: ' '
+    });
+
+    const description = txtgen.paragraph();
```
AlbertoPdRF: Maybe we could use the `article()` method with a random number of paragraphs here so we can have event descriptons with different lengths, which will allow us to check that everything looks great regardless of the size of them.

AlbertoPdRF (changes requested):
As always, great work @Be-Like! :100:

Please see some (mostly nitpicks) comments inline.

Regarding your questions/comments:
1. I'm not sure if I'd rather proceed with the clear warning and possibility of abortion path that we set with the interests collection instead of just dropping collections that the developer didn't ask for. This is something very tricky, as many collections are tightly coupled (e.g. should we also drop the events collection if the users one is dropped?)
2. That's a good assumption imho! However, as the `endDate` must be strictly greater than the `startDate`, it should range from `startDate` plus a minute (so the difference is visible in the UI) until `futureDate`. Please note that I didn't leave an inline comment regarding this.
3. That's a very good idea! Just for the sake of consistency, I think I would rename the file to `utils.js`.
4. That SGTM! See my inline comment about the formula to use to get that number right though.

And now some coming from me:
1. The files still have some styling issues :slightly_frowning_face:
2. The API's README needs some very small adjustments :slightly_smiling_face:
3. As with the users' avatars, it would be nice to add, at some point, cover photos to the events

I think that's all, thanks again for the contribution! 🤘

api/db/populate-events.js
Be-Like: Would you be opposed to each word having the first letter capitalized? There is a built in styling that the unique names generator provides.

api/db/populate-events.js
Be-Like: What is the longest a description can be? If you didn't provide one, there may be a restriction with mongodb that I can look up.

Be-Like: Just looked up mongodb's restrictions. The largest size a document can be is 16 MB, so we should be good.

api/db/populate-events.js
AlbertoPdRF: The "natural" way would probably be having only the first one capitalized, but feel free to go either way!

AlbertoPdRF: It feels to me that the "natural" way would be having only the first one capitalized, but feel free to go either way!

Be-Like:
> The files still have some styling issues 🙁

That is weird. When I commit my changes I see the linter running and changing my styling. 🤷‍♂️
> As with the users' avatars, it would be nice to add, at some point, cover photos to the events

I agree. Unless somebody else wants to look into this, I plan on looking into it once we get most of the other population scripts. 

With regards to generating the article, I set it to have a random amount of paragraphs between 0 and 14. I though that would be enough for our needs in front end development and still save some db space for those that are running the free MongoDb atlas - though I know they are pretty generous with their free version.

api/db/drop.js
```diff
@@ -5,6 +5,14 @@ const { connectDb, closeDb } = require('./connection');
 
 const Interest = require('../models/Interest');
 const User = require('../models/User');
+const Event = require('../models/Event');
+const Attendant = require('../models/Attendant');
+const { createPrompt } = require('./utils');
```
AlbertoPdRF: This should probably go on line 5 (and leaving a blank line before the models' imports).

api/db/drop.js
```diff
@@ -5,6 +5,14 @@ const { connectDb, closeDb } = require('./connection');
...
+
+const interestCount = async () => await Interest.countDocuments();
+const userCount = async () => await User.countDocuments();
+const eventCount = async () => await Event.countDocuments();
+const attendantCount = async () => await Attendant.countDocuments();
```
AlbertoPdRF:
```suggestion
const getInterestsCount = async () => await Interest.countDocuments();
const getUsersCount = async () => await User.countDocuments();
const getEventsCount = async () => await Event.countDocuments();
const getAttendantsCount = async () => await Attendant.countDocuments();
```
Following conventions, functions should start with verbs. Please note that I also pluralized the nouns.

⚠️ This will require changes elsewhere!

api/db/drop.js
```diff
@@ -25,18 +33,58 @@ const dropInterests = async () => {
...
     await User.collection.drop();
     console.log('\x1b[31m', 'Dropped users collection');
   } else {
     console.log('\x1b[33m', 'Users collection is already empty');
+    return;
+  }
+
+  if (await eventCount() !== 0) {
+    const drop = await createPrompt(
+      'The events collection is dependent on the events collection. Drop events collection?'
```
AlbertoPdRF:
```suggestion
      'The events collection is dependent on the users collection. Drop events collection?'
```
😛

api/db/drop.js
```diff
@@ -15,7 +23,7 @@ const userArgs = minimist(process.argv.slice(2), {
 
 const dropInterests = async () => {
```
AlbertoPdRF, line 24: The users and events collections depend on the interests one, so probably adding a prompt for each asking the developer if they want to drop them is a good idea.

api/db/drop.js
```diff
@@ -49,9 +97,15 @@ const dropCollection = async () => {
...
+    case 'events':
+      await dropEvents();
+      break;
+    case 'attendants':
+      await dropAttendants();
+      break;
     default:
       console.log(
-        `Unknown collection '${userArgs.c}', possible options are: [interests, users]`
+        `Unknown collection '${userArgs.c}', possible options are: [interests, users, events]`
```
AlbertoPdRF:
```suggestion
        `Unknown collection '${userArgs.c}', possible options are: [interests, users, events, attendants]`
```
Let's just add the attendants collection here too.

api/db/populate-events.js
```diff
@@ -0,0 +1,104 @@
...
+
+  for (let i = 0; i < 50 * multiplier; i++) {
+    const name = uniqueNamesGenerator({
+      dictionaries: [adjectives, animals, colors, starWars],
+      length: 3,
+      style: 'capital',
+      separator: ' '
+    });
+
+    const description = txtgen.article(Math.floor(Math.random() * 15));
```
AlbertoPdRF:
```suggestion
    const description = txtgen.article(Math.floor(Math.random() * 2) + 1);
```
Maybe 15 is too much, as that will look huge! And 0 would make the event invalid.

api/db/populate-events.js
```diff
@@ -0,0 +1,104 @@
...
+
+    const startDate = getRandomDate(now, futureDate);
+    const endDate = getRandomDate(startDate, futureDate);
```
AlbertoPdRF: Same comment about using `startDate` here as before: it should be at least a minute later. Also, maybe we can make things more realistic here and constraint the duration of events to anything since the said minute to day or two maximum? Feel free to go either way regarding the second part of the comment!

AlbertoPdRF (changes requested):
We are almost there! See my comments in line.

Regarding the code style, I still see some issues, ping me on Slack if you need help resolving those!

api/db/drop.js
```diff
@@ -23,27 +23,43 @@ const userArgs = minimist(process.argv.slice(2), {
...
   }
+
+  if (await getEventsCount() !== 0) {
+    const drop = await createPrompt(
+      'The events collection is dependent on the interests collection. Drop events collection?'
+    );
+
+    if (drop) await dropEvents(drop);
+  }
+
```
AlbertoPdRF: Upon further reflection, I think you can get rid of this one, as this question will already be asked if the developer decides to drop the users collection, and if not, they likely won't want to drop the events one either. Feel free to go either way!

api/db/populate-events.js
```diff
@@ -71,13 +72,16 @@ const createEvents = async (multiplier, randomLocation, verbose) => {
...
+      separator: ' ',
     });
 
-    const description = txtgen.article(Math.floor(Math.random() * 15));
+    const description = txtgen.article(Math.floor(Math.random() * 2) + 1);
 
     const startDate = getRandomDate(now, futureDate);
-    const endDate = getRandomDate(startDate, futureDate);
+    const endDate = getRandomDate(
+      startDate,
```
AlbertoPdRF:
```suggestion
      new Date(startDate.getTime() + 60000),
```

AlbertoPdRF (changes requested):
Two comments inline, and still the same with the linter issues 😢 Do you want me to push a commit with those fixes?

AlbertoPdRF (approved):
This looks great to me, thanks for taking another step towards completing #279! :shipit:

PR #331, opened 2020-10-08
Add script to populate attendants collection

Be-Like:
Ref #279.

### DIFF
--- a/api/db/populate-attendants.js
+++ b/api/db/populate-attendants.js
@@ -1,0 +1,49 @@
+const User = require('../models/User');
+const Event = require('../models/Event');
+const Attendant = require('../models/Attendant');
+
+let displayLogs;
+
+const createAttendant = async (attendantParams) => {
+  const attendant = new Attendant({
+    user: attendantParams.user,
+    event: attendantParams.event,
+  });
+  await attendant.save();
+
+  if (displayLogs) {
+    console.log('\n', '\x1b[0m', `New attendant created: ${attendant}`);
+  }
+};
+
+const createAttendants = async (verbose) => {
+  console.log('\n', '\x1b[0m', 'Populating attendants collection...');
+
+  let usersList = await User.distinct('_id');
+  if (usersList.length === 0) {
+    console.log(
+      '\x1b[33m',
+      'The users collection does not exist. Aborted attendants collection population.'
+    );
+    return;
+  }
+
+  let eventsList = await Event.find();
+  if (eventsList.length === 0) {
+    console.log(
+      '\x1b[33m',
+      'The events collection does not exist. Aborted attendants collection population.'
+    );
+    return;
+  }
+
+  let attendantsList = await Attendant.find();
+
+  // For each event, get a random number of users <= attendants limit
+  // Attendants limit - attendees
+  eventsList.forEach((event) => {
+    console.log('Event');
+  });
+};
+
+module.exports = { createAttendants };
--- a/api/db/populate-attendants.js
+++ b/api/db/populate-attendants.js
@@ -1,3 +1,5 @@
+const { getRandomSubset } = require('./utils');
+
 const User = require('../models/User');
 const Event = require('../models/Event');
 const Attendant = require('../models/Attendant');
@@ -6,8 +8,8 @@ let displayLogs;
 
 const createAttendant = async (attendantParams) => {
   const attendant = new Attendant({
-    user: attendantParams.user,
     event: attendantParams.event,
+    user: attendantParams.user,
   });
   await attendant.save();
 
@@ -37,13 +39,29 @@ const createAttendants = async (verbose) => {
     return;
   }
 
+  displayLogs = verbose;
   let attendantsList = await Attendant.find();
+  const attendantsArray = [];
 
-  // For each event, get a random number of users <= attendants limit
-  // Attendants limit - attendees
-  eventsList.forEach((event) => {
-    console.log('Event');
-  });
+  for (const event of eventsList) {
+    const attendees = attendantsList
+      .filter((attendee) => attendee.event.toString() == event._id.toString())
+      .map((attendee) => attendee.user.toString());
+
+    const users = usersList.filter(
+      (user) => !attendees.includes(user.toString())
+    );
+
+    const availability = event.attendantsLimit - attendees.length;
+    const subset = getRandomSubset(users, availability);
+
+    for (const user of subset) {
+      const attendantParams = { event, user };
+      attendantsArray.push(await createAttendant(attendantParams));
+    }
+  }
+
+  console.log('\x1b[32m', `Created ${attendantsArray.length} attendants`);
 };
 
 module.exports = { createAttendants };
--- a/api/db/populate-attendants.js
+++ b/api/db/populate-attendants.js
@@ -60,13 +60,13 @@ const createAttendants = async (verbose) => {
     const spotsLeft = event.attendantsLimit - attendees.length;
     if (spotsLeft <= 0 || !proceed) continue;
 
-    const potentialAttendees = usersList.filter(
-      (user) => !attendees.includes(user.toString())
-    );
-
     const eventCreator = event.createdBy.toString();
-    potentialAttendees = potentialAttendees.filter(
-      (user) => user !== eventCreator
+    const potentialAttendees = usersList.filter(
+      (user) =>
+        !(
+          attendees.includes(user.toString()) ||
+          user.toString() === eventCreator
+        )
     );
 
     const creatorIncluded = attendees.includes(eventCreator);
--- a/api/db/populate-attendants.js
+++ b/api/db/populate-attendants.js
@@ -64,12 +64,15 @@ const createAttendants = async (verbose) => {
       (user) => !attendees.includes(user.toString())
     );
 
-    const creatoreIncluded = attendees.includes(event.createdBy.toString());
+    const eventCreator = event.createdBy.toString();
+    const creatorIncluded = attendees.includes(eventCreator);
     const subset = getRandomSubset(
       potentialAttendees,
-      creatoreIncluded ? spotsLeft : spotsLeft - 1
+      creatorIncluded ? spotsLeft : spotsLeft - 1
     );
-    if (!creatoreIncluded) subset.unshift(event.createdBy.toString());
+    if (!creatorIncluded && !subset.includes(eventCreator)) {
+      subset.unshift(event.createdBy.toString());
+    }
 
     for (const user of subset) {
       const attendantParams = { event, user };
--- a/api/db/populate-attendants.js
+++ b/api/db/populate-attendants.js
@@ -54,11 +54,15 @@ const createAttendants = async (verbose) => {
 
   for (const event of eventsList) {
     const attendees = attendantsList
-      .filter((attendee) => attendee.event.toString() == event._id.toString())
+      .filter((attendee) => attendee.event.toString() === event._id.toString())
       .map((attendee) => attendee.user.toString());
 
     const spotsLeft = event.attendantsLimit - attendees.length;
-    if (spotsLeft <= 0 || !proceed) continue;
+    const proceedByLimit = event.attendantsLimit - 1 >= spotsLeft;
+    if (spotsLeft <= 0 || (!proceed && proceedByLimit)) {
+      console.log('Skipped for this event');
+      continue;
+    }
 
     const eventCreator = event.createdBy.toString();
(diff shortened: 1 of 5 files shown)


api/db/populate-attendants.js
```diff
@@ -0,0 +1,67 @@
...
+      'The events collection does not exist. Aborted attendants collection population.'
+    );
+    return;
+  }
+
+  displayLogs = verbose;
+  let attendantsList = await Attendant.find();
+  const attendantsArray = [];
+
+  for (const event of eventsList) {
```
AlbertoPdRF, line 55: In other places, I've been using the `forEach()` method instead. Maybe we should change it here too for consistency, but I don't feel too strongly about that.

api/db/populate-attendants.js
```diff
@@ -0,0 +1,67 @@
...
+    const attendees = attendantsList
```
AlbertoPdRF, line 56: Why not use `attendants` here (`attendant` when looping through it below) instead of `attendees` (`attendee`)?

api/db/populate-attendants.js
```diff
@@ -0,0 +1,67 @@
...
+      .filter((attendee) => attendee.event.toString() == event._id.toString())
+      .map((attendee) => attendee.user.toString());
+
+    const users = usersList.filter(
```
AlbertoPdRF: The name of this variable should make it clear that these are users who are not already attending the event. Maybe, to avoid complexity, we should skip events that already have attendants?

Also, the creator of the event should by default be listed as an attendant.

api/db/utils.js
```diff
@@ -25,15 +25,15 @@ const getRegion = (random, country) =>
     ? country.regions[Math.floor(Math.random() * country.regions.length)]
     : { shortCode: 'M' };
 
-const getRandomInterests = (interests) =>
-  interests
+const getRandomSubset = (list, maxSize) =>
```
AlbertoPdRF: This is nice! While we are at it, let's also add a `minSize` here (it can be 0 by default), so we can fix the issue mentioned on the last sentence of [this comment](https://github.com/Tisn/tisn.app/issues/279#issuecomment-705363801).

AlbertoPdRF (changes requested):
This is looking good @Be-Like! :sparkles: Just a couple of minor things to fix and we are ready to go!

api/db/populate-attendants.js
Be-Like: The only thing with `forEach` is that it cant be asynchronous (last time I checked). That's why I went with the `for..of` approach.

api/db/populate-attendants.js
AlbertoPdRF: And while we speak about events, I just realized that at some point we changed the default number of events created from `500` to `50`. Let's change it back to `500`.

api/db/populate-attendants.js
Be-Like: I can change it, though the plural of attendant is attendees. Attendants is the plural of a service worker (ie. flight attendants).

AlbertoPdRF: Oh, wow, then we should change it everywhere else instead! :open_mouth: Do you want to open an issue about this?

api/db/populate-attendants.js
```diff
@@ -0,0 +1,81 @@
...
+      'The events collection does not exist. Aborted attendants collection population.'
+    );
+    return;
+  }
+
+  displayLogs = verbose;
+  let attendantsList = await Attendant.find();
+  const attendantsArray = [];
+
+  let affirmativeResponse = false;
```
AlbertoPdRF: I would rename this variable to `proceed` or something similar, as it's done in other scripts, as the current name sounds a bit weird to me.

api/db/populate-attendants.js
```diff
@@ -0,0 +1,81 @@
...
+      'Some events already contain attendees. Would you like to add attendees to events that already contain attendees?'
+    );
+  }
+
+  for (const event of eventsList) {
+    const attendees = attendantsList
+      .filter((attendee) => attendee.event.toString() == event._id.toString())
+      .map((attendee) => attendee.user.toString());
+
+    const availability = event.attendantsLimit - attendees.length;
```
AlbertoPdRF: Maybe `availableSpots`, `spotsLeft`, or something similar would be a clearer name.

api/db/populate-attendants.js
```diff
@@ -0,0 +1,81 @@
...
+    if (!affirmativeResponse && event.attendantsLimit - 1 >= availability) {
```
AlbertoPdRF: I think the logic of this condition isn't catching all the cases where we want to skip adding attendees to an event. For instance, if `availability` is `0` the condition should evaluate to `true`, regardless of the value of `affirmativeResponse`.

Also, the `- 1` here should only be taken into account if the event creator isn't an attendee yet, no?

api/db/populate-attendants.js
```diff
@@ -0,0 +1,81 @@
...
+      continue;
+    }
+
+    const potentialAttendees = usersList.filter(
+      (user) => !attendees.includes(user.toString())
+    );
+
+    const subset = getRandomSubset(potentialAttendees, availability - 1);
```
AlbertoPdRF: I would rename this to make it clear what the subset is about.

And same comment as above about the `- 1` here: it should only be taken into account if the event creator isn't an attendee yet, no?

api/db/populate-attendants.js
```diff
@@ -0,0 +1,81 @@
...
+    subset.unshift(event.createdBy.toString());
```
AlbertoPdRF: And same here again: Also, the `- 1` the event creator should only be added to `subset` if they are not an attendee yet.

AlbertoPdRF (changes requested):
A few more comments from me! 🧐

api/db/populate-attendants.js
Be-Like: I fought with this one for a while. 😂 

I agree with what your saying on the latter point. However, it is such a convoluted  possibility that it is almost impractical to include when just populating attendees list. The attendees would essentially have to be populated, then a user would have to manually go in and unattend an event and the event would have to not have any other attendees, and then they would have to re-run the populate attendants script again. 

To handle such a situation would be expensive too. Plus ideally, once there are actual users, we wouldn't need the population scripts as much because you would probably want to have test data based off your actual production db.

Be-Like: That being said, I can try to figure it out. 😅

Be-Like: I figured it out. And of course it was super simple! 😅 Just had to take a break and come back to it.

api/db/populate-attendants.js
Be-Like: This one is correct. It is to compensate for the following line `subset.unshift(event.createdBy.toString());` which adds the event creator as an attendant.

api/db/populate-attendants.js
AlbertoPdRF: Hahaha, that's what normally happens! 🤣

api/db/populate-attendants.js
```diff
@@ -0,0 +1,85 @@
...
+
+  for (const event of eventsList) {
+    const attendees = attendantsList
+      .filter((attendee) => attendee.event.toString() == event._id.toString())
+      .map((attendee) => attendee.user.toString());
+
+    const spotsLeft = event.attendantsLimit - attendees.length;
+    // const proceedByLimit =  ?
+    //   event.attendantsLimit >= spotsLeft :
+    //   event.attendantsLimit - 1 >= spotsLeft;
```
AlbertoPdRF:
```suggestion
```
This is vestigial, right?

api/db/populate-attendants.js
```diff
@@ -0,0 +1,85 @@
...
+    const proceedByLimit = event.attendantsLimit - 1 >= spotsLeft;
+    if (proceedByLimit || (!proceed && proceedByLimit)) {
+      continue;
+    }
```
AlbertoPdRF:
```suggestion
    if (!proceed || spotsLeft === 0) continue;
```
I think rewriting the condition as suggested makes what's happening more clear. How do you see this suggestion?

api/db/populate-attendants.js
```diff
@@ -0,0 +1,85 @@
...
+
+    const potentialAttendees = usersList.filter(
+      (user) => !attendees.includes(user.toString())
+    );
+
+    const subset = getRandomSubset(potentialAttendees, spotsLeft - 1);
+    subset.unshift(event.createdBy.toString());
+
+    for (const user of subset) {
```
AlbertoPdRF:
```suggestion
    const creatorIncluded = attendees.includes(event.createdBy.toString());

    const newAttendees = getRandomSubset(
      potentialAttendees,
      creatorIncluded ? spotsLeft : spotsLeft - 1
    );
    if (!creatorIncluded) newAttendees.unshift(event.createdBy.toString());

    for (const user of newAttendees) {
```
This is what I meant with my earlier comment.

AlbertoPdRF (changes requested):
I left some more comments, all with suggestions so hopefully it's more clear what I meant. :warning: All the suggestions are untested.

api/db/populate-attendants.js
AlbertoPdRF: I left a new comment with a suggestion to try to make what I meant more clear. Let's discuss this there!

api/db/populate-attendants.js
AlbertoPdRF: And same here: I left a new comment with a suggestion to try to make what I meant more clear. Let's discuss this there!

api/db/populate-attendants.js
AlbertoPdRF: And another issue with this is that we have to make sure that `potentialAttendees` does not include the event creator, as, if it did, they could be included twice on `newAtendees`.

api/db/populate-attendants.js
Be-Like: Yeah, I'll be removing that.

api/db/populate-attendants.js
Be-Like: I did change this to ` if (!proceed || spotsLeft <= 0) continue;`. I know this shouldn't ever happen, however it does handle the possibility of having less than 0 spots left.

api/db/populate-attendants.js
```diff
@@ -0,0 +1,83 @@
...
+    const potentialAttendees = usersList.filter(
+      (user) => !attendees.includes(user.toString())
+    );
+
+    const creatoreIncluded = attendees.includes(event.createdBy.toString());
+    const subset = getRandomSubset(
+      potentialAttendees,
+      creatoreIncluded ? spotsLeft : spotsLeft - 1
+    );
+    if (!creatoreIncluded) subset.unshift(event.createdBy.toString());
```
AlbertoPdRF: As this is now, we could have the creator included twice `subset`, right? When `creatorIncluded` is `false`, we should the creator them from `potentialAttendees` before getting the random subset to avoid that.

Also, shouldn't this be `creatorIncluded`, without the extra `e`? And while we talk about names, `subset` could also use a better name, but feel free to leave it as is!

AlbertoPdRF (changes requested):
Just one last comment before this is ready to be shipped! :ship:

api/db/populate-attendants.js
Be-Like: Yeah, adding the `e` was a typo.

api/db/populate-attendants.js
```diff
@@ -0,0 +1,91 @@
...
+    if (spotsLeft <= 0 || !proceed) continue;
+
+    const potentialAttendees = usersList.filter(
+      (user) => !attendees.includes(user.toString())
+    );
+
+    const eventCreator = event.createdBy.toString();
+    potentialAttendees = potentialAttendees.filter(
+      (user) => user !== eventCreator
+    );
```
AlbertoPdRF:
```suggestion
    const eventCreator = event.createdBy.toString();
    const potentialAttendees = usersList.filter(
      (user) =>
        !(
          attendees.includes(user.toString()) ||
          user.toString() === eventCreator
        )
    );
```
`potentialAttendees` can't be reasigned, as it's `const`, I've made a suggestion to do it all at once instead :rocket: If you agree with my suggestion, please double check that it works before commiting it, as this is a bit tricky!

AlbertoPdRF (changes requested):
There's an error in the code, please see the comment in line.

AlbertoPdRF (approved):
This LGTM! Thanks for the contribution and for your infinite patience debugging this with together! :shipit:

PR #334, opened 2020-10-09
Implement database backup mechanism

### LINKED ISSUE #27: Create a daily back up of the database.
Losing that information would be a total disaster!

jacklagare:
Resolves #27 

- Implemented function for creating a database dump
- Implemented a cron job and initialized it upon application
- Implemented function for sending the database dump to an admin email


api/.env-template
```diff
@@ -5,3 +5,4 @@ CLOUDINARY_API_KEY=""
 CLOUDINARY_API_SECRET=""
 SENDGRID_API_KEY="SG."
 BASE_CLIENT_URL="http://localhost:3000"
+ADMIN_EMAIL=""
```
AlbertoPdRF: I think I'd rather hard-code this in the code to be `admin@tisn.app` than have it here, at least for now.

api/app.js
```diff
@@ -47,4 +47,6 @@ app.use(function (err, req, res, next) {
   return;
 });
 
+global.CronJob = require('./utils/cron.js');
+
```
AlbertoPdRF, line 51: Is this necessary? I'm checking cron's README and can't find a reference to it. (I've never used this package before, so sorry if this is a stupid question!)

api/db/backup.js
```diff
@@ -0,0 +1,47 @@
+const exec = require('child_process').exec;
+
+// Retrieve database connection parameters
+const getDbUrl = () => {
+  require('dotenv').config();
+  return process.env.DB_URL;
+};
+
```
AlbertoPdRF: This function already exists in the `connection.js` file, it would be better if you added it to the exports there and just import it here.

api/db/backup.js
```diff
@@ -0,0 +1,47 @@
...
+const createBackup = () => {
+  // Prepare database archive filename
+  currentDate = new Date();
+
+  let databaseDumpFilename =
```
AlbertoPdRF: A few suggestions here:
- Use `const` instead of `let`
- Remove `database` from the name
- Use template literals/strings
- As the file will go inside `db/dumps`, maybe we can just name it `yearmonthdate.gz`
  - Please create the `db/dumps` folder with an empty file inside it called `.keep` (as it's done with the `uploads` folder)

api/db/backup.js
```diff
@@ -0,0 +1,47 @@
...
+    'tisn_db_dump_' +
+    currentDate.getFullYear() +
+    '_' +
+    (currentDate.getMonth() + 1) +
+    '_' +
+    currentDate.getDate() +
+    '.gz';
+
+  // Prepare mongodump command
+  let dumpCmd =
```
AlbertoPdRF: Could you elaborate on how you constructed the command? I have to check it when I get back home, but I think I use some different and/or extra arguments.

Also, same here regarding `const` instead of `let` and the use of template literals/strings.

api/db/backup.js
```diff
@@ -0,0 +1,47 @@
...
+    getDbUrl() +
+    "' --archive=dump/" +
+    databaseDumpFilename +
+    ' --gzip';
+
+  // Execute mongodump command
+  exec(dumpCmd, (error, stdout, stderr) => {
+    if (error) {
+      console.log(error);
+      return 0;
```
AlbertoPdRF: Do we need the `return` here?

api/package.json
```diff
@@ -22,12 +22,14 @@
...
     "cors": "^2.8.5",
     "country-region-data": "^1.6.0",
+    "cron": "^1.8.2",
     "debug": "~4.2.0",
     "dotenv": "^8.2.0",
     "express": "~4.17.1",
     "express-fileupload": "^1.2.0",
     "express-jwt": "^6.0.0",
     "express-validator": "^6.6.1",
+    "fs": "0.0.1-security",
```
AlbertoPdRF: I think this shouldn't be added, see [here](https://www.npmjs.com/package/fs).

api/utils/cron.js
```diff
@@ -0,0 +1,9 @@
+const CronJob = require('cron').CronJob;
+const DatabaseBackup = require('../db/backup.js');
```
AlbertoPdRF: For consistency with the latest additions to the code, this should be `const { createBackup } = require('../db/backup.js');`

api/utils/cron.js
```diff
@@ -0,0 +1,9 @@
...
+
+// Configure cron job to run every day at 12:00am
+let job = new CronJob('00 00 00 * * *', function () {
```
AlbertoPdRF: Can this be a `const`, or is it required to be `let`? Also, we use arrow functions everywhere, so it would be better to use one for the callback too.

api/utils/cron.js
```diff
@@ -0,0 +1,9 @@
...
+  DatabaseBackup.createBackup();
+});
+
+job.start();
```
AlbertoPdRF: Can we only start the cron job in production? I don't think we want this running on any other environment.

api/utils/emails.js
```diff
@@ -1,11 +1,69 @@
...
 const Token = require('../models/Token');
-
 const crypto = require('crypto');
 const sendgrid = require('@sendgrid/mail');
+const fs = require('fs');
+const path = require('path');
 
 sendgrid.setApiKey(process.env.SENDGRID_API_KEY);
 
+const emailDatabaseBackup = (filename) => {
```
AlbertoPdRF: I think it would be better to name this `dbBackup`.

api/utils/emails.js
```diff
@@ -1,11 +1,69 @@
...
+  try {
+    let attachment = fs
```
AlbertoPdRF: A few questions here:
- Can this be `const` instead of `let`?
- Why use `readFileSync` instead of `readFile`?
- We use `path.join` elsewhere, can this be changed for consistency?
- I guess the `toString('base64')` is required by SendGrid, right?

api/utils/emails.js
```diff
@@ -1,11 +1,69 @@
...
+    // Ensure that the email address is set
+    if (adminEmail == undefined || adminEmail == '') {
+      return 0;
+    }
+
+    // Create message
+    let message = {
+      to: adminEmail,
+      from: { email: 'no-reply@tisn.app', name: 'Tisn' },
+      subject: 'TISN APP - Database Backup',
```
AlbertoPdRF: Let's just use `Database backup` here.

api/utils/emails.js
```diff
@@ -1,11 +1,69 @@
...
+      text: 'Attached is the database dump ' + filename + '.',
```
AlbertoPdRF: Same here about template literals/strings.

api/utils/emails.js
```diff
@@ -1,11 +1,69 @@
...
+        {
+          content: attachment,
+          filename: filename,
+          type: 'application/gzip',
+          disposition: 'attachment',
+        },
+      ],
+    };
+
+    console.log('Now sending email...');
```
AlbertoPdRF: I think I would remove all the unnecessary `console.log` from the file.

AlbertoPdRF (changes requested):
Wow, such an amazing first contribution @jacklagare! :astonished:

I left a bunch of comments inline, many of them because almost all of this is new to me and to make sure that everything is as intended.

Also, do you have any suggestion on how to test this before pushing it to production?

Thanks so much for taking the time to contribute! :100:

AlbertoPdRF:
Oh, and one other small change to make: with these changes, the high level `.gitignore` file can be removed. Instead, something like

```
# dumps
/db/dumps/*
```

should be added to the API's folder one (just like it's done with the uploads folder). This makes me wonder... should we delete the dump from the server after the email has been sent? We already do that when uploading files.

api/app.js
jacklagare: This is the line of code that makes sure that the function is running as well when the API is started.

api/utils/emails.js
jacklagare: 1. Replaced with const
2. You're right, it should be readFile. Changed that already.
3. Done
4. Yes. The attachment should be encoded to base64. I updated the code for this so it's now more readble.

jacklagare:
@AlbertoPdRF I already made the necessary changes.

AlbertoPdRF:
Thanks for the quick iteration @jacklagare! :fast_forward: To avoid providing another lenghty review, I've pushed [a commit](https://github.com/AlbertoPdRF/tisn.app/commit/4f3865b93b3650726cf2ebb185f8a46ca94660dc) to my fork with all the changes I would do. If you agree with them, feel free to cherry-pick that commit and push it to your branch, or I can do that myself if you want me to.

Apart from that, I still need to test this thoroughly to make sure that it will work when pushed to production. :smile:

jacklagare:
@AlbertoPdRF, I cherry-picked your commits.

I removed `--oplog` from the mongodump command though as it causes an error.

AlbertoPdRF:
I'll try setting up an staging environment so we can test this on a production-like environment later today!

jacklagare:
Noted @AlbertoPdRF. Thanks!

AlbertoPdRF (approved):
I've tested this and it works like a charm! Thanks so much for your contribution to Tisn @jacklagare 🎉

A few notes just to keep them written somewhere:
- I had to install the `http://github.com/uhray/heroku-buildpack-mongo.git` buildpack on Heroku (in addition to the `heroku/nodejs` one)
- I'm concerned about not being able to use the `--oplog` argument for the `mongodump` command, as this may make us lose information in case we actually have to `mongorestore` the dump at some point. I guess that if we grow bigger, we'll have to investigate other options for the backups

Anyway, this is a GREAT addition to the code base, so I'm merging it right now! 🚢

PR #335, opened 2020-10-09
Add scripts to populate and drop the friendships collection

Be-Like:
Refer #279.

### DIFF
--- a/api/db/populate-friendships.js
+++ b/api/db/populate-friendships.js
@@ -11,7 +11,6 @@ const createFriendship = async (friendshipParams) => {
     receivant: friendshipParams.receivant,
     accepted: friendshipParams.accepted,
     acceptedAt: friendshipParams.acceptedAt,
-    lastMessageAt: friendshipParams.lastMessageAt,
   });
   await friendship.save();
 
@@ -37,12 +36,12 @@ const createFriendships = async (verbose) => {
   }
 
   for (const user of usersList) {
-    const containFriends = friendshipsArray.some(
+    const alreadyHasFriends = friendshipsArray.some(
       (friendship) =>
         friendship.requestant.toString() === user.toString() ||
         friendship.receivant.toString() === user.toString()
     );
-    if (!proceed && containFriends) continue;
+    if (!proceed && alreadyHasFriends) continue;
 
     // split current user from all others
     const tmpList = usersList.filter((userValue) => userValue !== user);
@@ -50,26 +49,22 @@ const createFriendships = async (verbose) => {
     const usersSubset = getRandomSubset(tmpList, tmpList.length + 1);
 
     for (const receivant of usersSubset) {
-      const alreadyFriends = friendshipsArray.some(
+      const alreadyAreFriends = friendshipsArray.some(
         (friendship) =>
           (friendship.requestant.toString() === user.toString() &&
             friendship.receivant.toString() === receivant.toString()) ||
           (friendship.requestant.toString() === receivant.toString() &&
             friendship.receivant.toString() === user.toString())
       );
-      if (alreadyFriends) continue;
+      if (alreadyAreFriends) continue;
 
       const now = new Date();
-      const acceptedAt = getRandomDate(
-        new Date(2020, 4, 4),
-        new Date(now.getTime() - 60000)
-      );
+      const acceptedAt = getRandomDate(new Date(2020, 4, 5), now);
       const friendshipParams = {
         requestant: user,
         receivant,
         accepted: Math.random() > 0.2,
         acceptedAt,
-        lastMessageAt: getRandomDate(acceptedAt, now),
       };
 
       friendshipsArray.push(await createFriendship(friendshipParams));
--- a/api/db/populate-friendships.js
+++ b/api/db/populate-friendships.js
@@ -27,12 +27,12 @@ const createFriendships = async (verbose) => {
 
   let usersList = await User.distinct('_id');
   const friendshipsArray = await Friendship.find();
-  let recordCount = friendshipsArray.length;
+  const documentCount = friendshipsArray.length;
 
   let proceed = true;
-  if (recordCount > 0) {
+  if (documentCount > 0) {
     proceed = await createPrompt(
-      'Some users already have friends. Would you like to add friends to users that already contain friends?'
+      'Some users already have friends. Would you like to add friends to users that already have friends?'
     );
   }
 
@@ -45,20 +45,19 @@ const createFriendships = async (verbose) => {
     if (!proceed && containFriends) continue;
 
     // split current user from all others
-    const tmpList = usersList.filter((userValue) => userValue != user);
+    const tmpList = usersList.filter((userValue) => userValue !== user);
     // create subset userList from remaining usersList
-    const usersSubset = getRandomSubset(tmpList, tmpList.length);
+    const usersSubset = getRandomSubset(tmpList, tmpList.length + 1);
 
     for (const receivant of usersSubset) {
-      const alreadFriends = friendshipsArray.filter((friendship) => {
-        return (
+      const alreadyFriends = friendshipsArray.some(
+        (friendship) =>
           (friendship.requestant.toString() === user.toString() &&
             friendship.receivant.toString() === receivant.toString()) ||
           (friendship.requestant.toString() === receivant.toString() &&
             friendship.receivant.toString() === user.toString())
-        );
-      });
-      if (alreadFriends.length > 0) continue;
+      );
+      if (alreadyFriends) continue;
 
       const now = new Date();
       const acceptedAt = getRandomDate(
@@ -68,7 +67,7 @@ const createFriendships = async (verbose) => {
       const friendshipParams = {
         requestant: user,
         receivant,
-        accepted: Math.random() > 0.1 ? true : false,
+        accepted: Math.random() > 0.2,
         acceptedAt,
         lastMessageAt: getRandomDate(acceptedAt, now),
       };
@@ -78,7 +77,7 @@ const createFriendships = async (verbose) => {
   }
   console.log(
     '\x1b[32m',
-    `Created ${friendshipsArray.length - recordCount} friendships`
+    `Created ${friendshipsArray.length - documentCount} friendships`
   );
 };
 
--- a/api/db/populate-friendships.js
+++ b/api/db/populate-friendships.js
@@ -1,0 +1,23 @@
+const User = require('../models/User');
+const Friendship = require('../models/Friendship');
+
+let displayLogs;
+
+const createFriendship = async (friendshipParams) => {
+  const friendship = new Friendship({
+    requestant: friendshipParams.requestant,
+    receivant: friendshipParams.receivant,
+    accepted: friendshipParams.accepted,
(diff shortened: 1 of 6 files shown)


api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+
+let displayLogs;
+
+const createFriendship = async (friendshipParams) => {
+  const friendship = new Friendship({
+    requestant: friendshipParams.requestant,
+    receivant: friendshipParams.receivant,
+    accepted: friendshipParams.accepted,
+    acceptedAt: friendshipParams.acceptedAt,
+    lastMessageAt: friendshipParams.lastMessageAt,
```
AlbertoPdRF: I think we should set this when populating the messages collection instead, but maybe I'm not seeing your intention here.

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+  return friendship;
+};
+
+const createFriendships = async (verbose) => {
+  console.log('\n', '\x1b[0m', 'Populating friendships collection...');
+  displayLogs = verbose;
+
+  let usersList = await User.distinct('_id');
+  const friendshipsArray = await Friendship.find();
```
AlbertoPdRF, line 28: For consistency, shouldn't this be named `friendshipsList`?

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+  let recordCount = friendshipsArray.length;
```
AlbertoPdRF: This should be `const` instead of `let`. Also, maybe name it `documentsCount`?

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+
+  let proceed = true;
+  if (recordCount > 0) {
+    proceed = await createPrompt(
+      'Some users already have friends. Would you like to add friends to users that already contain friends?'
```
AlbertoPdRF: Nitpick: I would also use `have friends` at the end of the sentence instead of `contain friends`.

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+    );
+  }
+
+  for (const user of usersList) {
+    const containFriends = friendshipsArray.some(
```
AlbertoPdRF: This name isn't following conventions, as it should not start with a verb, although I can't really come up with a name that follows them right now 😕

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+      (friendship) =>
+        friendship.requestant.toString() === user.toString() ||
+        friendship.receivant.toString() === user.toString()
+    );
+    if (!proceed && containFriends) continue;
+
+    // split current user from all others
+    const tmpList = usersList.filter((userValue) => userValue != user);
```
AlbertoPdRF: Why not use the `toString` method here too and use `!==` instead of `!=`?

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+    // create subset userList from remaining usersList
+    const usersSubset = getRandomSubset(tmpList, tmpList.length);
```
AlbertoPdRF: Shouldn't the second parameter be `tmpList.length + 1`?

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+
+    for (const receivant of usersSubset) {
+      const alreadFriends = friendshipsArray.filter((friendship) => {
```
AlbertoPdRF: I think this should be a `some` instead of a `filter`. Also, you can remove the `{ return (  ) }` part and only leave everything in between. Oh, and a nitpick: you've missed the `y` of `already`.

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+            friendship.receivant.toString() === receivant.toString()) ||
+          (friendship.requestant.toString() === receivant.toString() &&
+            friendship.receivant.toString() === user.toString())
+        );
+      });
+      if (alreadFriends.length > 0) continue;
+
+      const now = new Date();
+      const acceptedAt = getRandomDate(
+        new Date(2020, 4, 4),
```
AlbertoPdRF: Nitpick: we are using May 5th, 2020 elsewhere.

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+        new Date(now.getTime() - 60000)
```
AlbertoPdRF: I'm curious, why the `- 60000` here?

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+      );
+      const friendshipParams = {
+        requestant: user,
+        receivant,
+        accepted: Math.random() > 0.1 ? true : false,
```
AlbertoPdRF: You can remove the ` ? true : false` part from here, as `Math.random() > 0.1` will already evaluate to `true` or `false`. Also, should we make it a bit more probable for a friendship request to not be accepted? Something like `Math.random >= 0.2` or `Math.random >= 0.25`?

api/db/populate-friendships.js
```diff
@@ -0,0 +1,85 @@
...
+        acceptedAt,
+        lastMessageAt: getRandomDate(acceptedAt, now),
+      };
+
+      friendshipsArray.push(await createFriendship(friendshipParams));
```
AlbertoPdRF, line 56: I would use an empty array for this, as we do in all other scripts.

api/db/populate-friendships.js
AlbertoPdRF: Also, as per one suggestion I've made below, this variable would only be used one time, so we can actually get rid of it and just evaluate it directly where it will be used.

AlbertoPdRF (changes requested):
Solid work @Be-Like!! 💯 Just a few hopefully easy to resolve details and we are good to go!

api/db/populate-friendships.js
Be-Like: Do you mean instead of having it be it's own script command?

AlbertoPdRF: Let me try to explain myself better:

When a friendship is created or accepted, the `lastMessageAt` value doesn't get set, it only gets set when a message between both users is actually sent. Thus, I think we should handle setting this value when populating the messages collection.

Be-Like: Gotcha, that makes sense. Are you opposed to keeping it as is until I work on the messages collection script?

api/db/populate-friendships.js
Be-Like: Yes and no. In order to populate the friendships collection, I need to get the friendships collection that already exists. And then as I create the friendships, I add them to this array so that I can check to see if the friendship exists in the next iteration. This should answer your last comment below as well.

Be-Like: Also, for all the other scripts it might be worth considering getting rid of the <collection>Array anyway. We don't use it for anything other console logging the amount of records generated, and that could just as easily be changed to an integer variable that we update whenever a record is saved to the db.

api/db/populate-friendships.js
Be-Like: I agree that it can be changed to a `const` and name change, but this needs to be calculated before we start adding records to the array. It is used to calculate the number of records generated. Calculates `end array length - before array length`. See comment above about needing to use the `friendshipsArray` to avoid duplicate documents.

api/db/populate-friendships.js
Be-Like: Sorry, I come from a Java and Ruby on Rails background and often type just the two symbols when evaluating. It's a typo. I'll change it, but we shouldn't need the `toString()` because we are evaluating a user object with another user object and not just their `id`'s.

api/db/populate-friendships.js
Be-Like: Would you like me to change it? Or remove duplicate code?

api/db/populate-friendships.js
Be-Like: Because I didn't know exactly what the `lastMessageAt` field was used for, so I wanted it to be somewhere in between the `acceptedAt` date and `now`. And I wanted there to be some possible variation between `acceptedAt` and `lastMessageAt` dates.

api/db/populate-friendships.js
Be-Like: See above comments. Empty array would make it much more complicated in trying to avoid duplicate friendships.

api/db/populate-friendships.js
AlbertoPdRF: I think I would remove it, as this as is can be misleading for developers visiting the `/chats` route, although seeing how quickly you are adding all scripts this won't be a problem for too long 😛

api/db/populate-friendships.js
AlbertoPdRF: Aha, that makes sense. And your second point makes sense too! I would say that, after all scripts are added, we should submit a follow up PR addressing this and other small issues that we might find to finally close #279.

api/db/populate-friendships.js
AlbertoPdRF: Yeah, let's change it, just for the sake of it.

api/db/populate-friendships.js
AlbertoPdRF: Aha, I see your reasoning now. I think we should be safe using `now` here, as we'll for sure have enough friendships with enough "room for messages", but feel free to keep this, extend it, or get rid of it!

api/db/populate-friendships.js
```diff
@@ -0,0 +1,84 @@
...
+  return friendship;
+};
+
+const createFriendships = async (verbose) => {
+  console.log('\n', '\x1b[0m', 'Populating friendships collection...');
+  displayLogs = verbose;
+
+  let usersList = await User.distinct('_id');
+  const friendshipsArray = await Friendship.find();
+  const documentCount = friendshipsArray.length;
```
AlbertoPdRF: Should be `documentsCount`?

api/db/populate-friendships.js
```diff
@@ -0,0 +1,84 @@
...
+
+  let proceed = true;
+  if (documentCount > 0) {
+    proceed = await createPrompt(
+      'Some users already have friends. Would you like to add friends to users that already have friends?'
+    );
+  }
+
+  for (const user of usersList) {
+    const containFriends = friendshipsArray.some(
```
AlbertoPdRF: Maybe change it to `alreadyHasFriends`? Or `Contains` instead of `Has`. The `alreadyFriends` below could be changed to `alreadyAreFriends` following the same pattern.

AlbertoPdRF (commented):
Apart from the replies to earlier comments, just two nitpicks!

api/db/populate-friendships.js
AlbertoPdRF: After giving this a second thought, maybe it will be better to remove it to avoid finding unnecessary constraints when populating the messages collection?

AlbertoPdRF (approved):
LGTM! Thanks for adding yet another awesome population script @Be-Like! 🔥

PR #336, opened 2020-10-10
Fix for incorrect dob format #311

### LINKED ISSUE #311: Profile Date of Birth Incorrect
**Describe the bug**
The date of birth displayed in the UserDetails.js is off by a day. The DB stores it correctly, so it is only in how the date is displayed. (Dates are the worst to handle in JS).

As an aside, this [stackoverflow article](https://stackoverflow.com/a/31732581/11357799) provides an excellent example of all the ways JS handles dates. 

My strong recommendation would be to utilize a library such as Moment.js or date-fns to handle your dates. It will handle the localization and time zones much better. 

**To Reproduce**

1. Go to Users
2. See error

**Expected behavior**
Date to be correct. 

**Screenshots**
image
image

idiglove:
Fixes #311

### DIFF
--- a/client/src/Components/UserDetails/UserDetails.js
+++ b/client/src/Components/UserDetails/UserDetails.js
@@ -15,7 +15,7 @@ import ChatIcon from '@material-ui/icons/Chat';
 
 import countries from 'country-region-data';
 
-import { distanceToNow, formatDateOnlyToString } from '../../logic/date-time';
+import { distanceToNow, formatDate } from '../../logic/date-time';
 
 import ShareMenu from '../ShareMenu/ShareMenu';
 
@@ -91,7 +91,7 @@ const UserDetails = (props) => {
         {restrictedDisplay && (
           <Fragment>
             <Typography variant="body1">
-              {formatDateOnlyToString(user.dateOfBirth, 'MM/dd/yyyy')}
+              {formatDate(user.dateOfBirth)}
             </Typography>
             <Typography gutterBottom variant="body1">
               {user.email}
--- a/client/src/Components/UserDetails/UserDetails.js
+++ b/client/src/Components/UserDetails/UserDetails.js
@@ -15,7 +15,7 @@ import ChatIcon from '@material-ui/icons/Chat';
 
 import countries from 'country-region-data';
 
-import { distanceToNow, formatDate } from '../../logic/date-time';
+import { distanceToNow, formatDateOnlyToString } from '../../logic/date-time';
 
 import ShareMenu from '../ShareMenu/ShareMenu';
 
@@ -91,7 +91,7 @@ const UserDetails = (props) => {
         {restrictedDisplay && (
           <Fragment>
             <Typography variant="body1">
-              {formatDate(user.dateOfBirth)}
+              {formatDateOnlyToString(user.dateOfBirth, 'MM/dd/yyyy')}
             </Typography>
             <Typography gutterBottom variant="body1">
               {user.email}
--- a/client/src/logic/date-time.js
+++ b/client/src/logic/date-time.js
@@ -24,13 +24,8 @@ const formatUtcToTimeZone = (dateString, formatString) =>
 export const formatInputDate = (dateString) =>
   formatUtcToTimeZone(dateString, 'yyyy-MM-dd');
 
-export const formatDate = (dateString) => formatUtcToTimeZone(dateString, 'P');
-
-//ignoring the timezone
-export const formatDateOnlyToString = (dateString, dateFormat) => {
-  const dateArray = dateString.substr(0, 10).split('-');
-  return dateArray[1] + '/' + dateArray[2] + '/' + dateArray[0];
-};
+export const formatDate = (dateString) =>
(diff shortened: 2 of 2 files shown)


AlbertoPdRF (changes requested):
First of all, thanks for your first contribution to Tisn @idiglove! 👏

Regarding the actual changes, there's some things to I'd like for us to do differently:
- As the `formatDate` function is only used once throughout the codebase, I think we should just edit it to call a new function different from `formatUtcToTimeZone` instead of defining a new one.
- The different function to call, let's name it `formatUtc` or something like that, should be pretty similar to `formatUtCToTimeZone`, but ignoring the timezone part, something like:

```javascript
const formatUtc = (dateString, formatString) =>
  format(parseISO(dateString), formatString, {
    locale: getLocale(),
  });
```
could work, although I haven't tested it! Please note that I'm making this suggestion because, althouh the date should not be time-zoned, it should definitely be localized.

Thanks again for contributing to Tisn! And if you have any doubt, feel free to ping me on Slack! 🙂

AlbertoPdRF:
As the suggested `formatUtc` function would only be used once, you can also merge that one with the `formatDate` one instead, whatever you prefer!

Oh, and I forgot about this before: can you include `Fixes #311` on your original comment so, when the PR is merged, the issue gets automatically closed? Thanks!

idiglove:
Hi @AlbertoPdRF 
I have tried something like that, but it doesnt work.

And now I have tried your suggestion,
```
const formatUtc = (dateString, formatString) =>
format(
  parseISO(dateString),
  formatString,
  {
    locale: getLocale(),
  }
);

export const formatInputDate = (dateString) =>
  formatUtcToTimeZone(dateString, 'yyyy-MM-dd');

export const formatDate = (dateString) => formatUtc(dateString, 'P');
```
It still doesnt work. I think parsing the datetime with the "T:00:..." will always make it change based on the user's timezone 

image

AlbertoPdRF:
Hm, that's interesting. In that case, could you try passing `dateString.split('T')[0]` to the `parseISO` function instead?

idiglove:
I have a new change now, what do you think? @AlbertoPdRF

AlbertoPdRF:
This LGTM! Just to double-check: @Be-Like could you too kindly confirm that this fixes the bug as you saw it? 🙂

Be-Like:
Looks like there is still an issue. This should read 04/02/1995. 
image

For this, I don't think you will be able to fully rely on the populate users script. That provides a date utilizing the utc. When a user signs up via the form, it sets the date correctly but sets the UTC to 00:00:00.

The picture below illustrates what I mean. Top document is for a user populated by script, while bottom is what is entered via the sign up form:
image

AlbertoPdRF:
Just a quick comment: we should make sure that the users' population script mimics what the sign up form does, so this is one of the issues that we should fix on the follow up PR I was referring to [here](https://github.com/Tisn/tisn.app/pull/335#discussion_r503351121).

idiglove:
Seems like my first fix will solve this? Just parse it like a string

AlbertoPdRF:
The problem in the case of your first fix is that we won't be getting a localized string date, which can be very misleading for the users.

idiglove:
> The problem in the case of your first fix is that we won't be getting a localized string date, which can be very misleading for the users.

What is an example of localized string date?

AlbertoPdRF:
In @Be-Like's case, for instance, the date is presented as `04/01/1995`, because it is localized in the `enUS` locale. If someone from Spain looked at that date, they would see `01/04/1995`, because it would be localized in the `es` locale. Now we only have two locales, so your suggested fix would be pretty easy to patch, but it's a better long-term fix to figure out how to use the library correctly so that this gets handled automatically.

idiglove:
> In @Be-Like's case, for instance, the date is presented as `04/01/1995`, because it is localized in the `enUS` locale. If someone from Spain looked at that date, they would see `01/04/1995`, because it would be localized in the `es` locale. Now we only have two locales, so your suggested fix would be pretty easy to patch, but it's a better long-term fix to figure out how to use the library correctly so that this gets handled automatically.

Ah I see. It's about the formatting, lemme try something that parses the date based on user's locale

AlbertoPdRF:
Did you get to try a solution like:
```javascript
export const formatDate = (dateString) =>
  format(parseISO(dateString.split('T')[0]), 'P', {
    locale: getLocale(),
  });
```
?

I feel like there has to be an easy solution to this that we are overlooking :stuck_out_tongue:

idiglove:
@Be-Like can you try again with my new fix? I dont know how to replicate your issue :)

Be-Like:
Looking good! Good work! I know how frustrating it can be to work with dates. They are the bane of every project I've ever worked on. 😅

image

image

idiglove:
Yay! Thanks for your help @AlbertoPdRF !

AlbertoPdRF (approved):
LGTM, congratulations @idiglove for your first contribution to Tisn! :tada:

And thanks @Be-Like for confirming that this resolves the bug :mag:

PR #337, opened 2020-10-11
Implement single origin validation through CORS 

### LINKED ISSUE #22: Restrict API to interact with the client only.
At least until it's crystal-clear that undesired actions can't be accomplished by interacting programatically with the API, its usage should be restricted to the client only.

By the way, this won't be needed while the app only has dummy data.

jacklagare:
Configured CORS so that only requests coming from `BASE_CLIENT_URL` are allowed. This will resolve #22.

### DIFF
--- a/api/app.js
+++ b/api/app.js
@@ -26,7 +26,20 @@ connectDb(process.env.DB_URL);
 
 require('./config/passport');
 
-app.use(cors());
+// Configure CORS for a single origin
+const corsOptions = {
+  origin: function (origin, callback) {
+    // If origin is the same as the value in BASE_CLIENT_URL, allow access
+    if (origin === process.env.BASE_CLIENT_URL) {
+      callback(null, true)
+    }
+    else {
+      callback(createError(403))
+    }
+  }
+}
+
+app.use(cors(corsOptions));
 app.use(logger('dev'));
 app.use(express.json());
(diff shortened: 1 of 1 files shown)


jacklagare:
Hi @AlbertoPdRF, here is the PR.

api/app.js
```diff
@@ -26,7 +26,20 @@ connectDb(process.env.DB_URL);
...
+  origin: function (origin, callback) {
+    // If origin is the same as the value in BASE_CLIENT_URL, allow access
+    if (origin === process.env.BASE_CLIENT_URL) {
+      callback(null, true)
+    }
+    else {
+      callback(createError(403))
+    }
+  }
+}
```
AlbertoPdRF:
```suggestion
const corsOptions = {
  origin: (origin, callback) => {
    if (origin === process.env.BASE_CLIENT_URL) {
      callback(null, true);
    } else {
      callback(createError(403));
    }
  },
};
```
This is just a code style suggestion

AlbertoPdRF (changes requested):
Nice and quick work @jacklagare! 🔝

Just a few code style issues to resolve.

jacklagare:
@AlbertoPdRF, changes have been committed.

AlbertoPdRF (approved):
LGTM! Thanks for the contribution @jacklagare! 📈

PR #338, opened 2020-10-12
Scripts to populate and drop comments collections

Be-Like:
Refer #279.


api/db/populate-comments.js
```diff
@@ -0,0 +1,94 @@
...
+  const commentsArray = [];
+  for (const event of eventsList) {
+    // Find attendants for event
+    const eventAttendants = attendantsList.filter(
+      (attendant) => attendant.event.toString() === event.toString()
+    );
+    const eventComments = await Comment.find({ event });
+
+    // Number of comments to generate
+    const commentCount = Math.floor(Math.random() * 20);
```
AlbertoPdRF: Nitpick: `commentsCount`

api/db/populate-comments.js
```diff
@@ -0,0 +1,94 @@
...
+    for (let i = 0; i <= commentCount; i++) {
+      const user =
+        eventAttendants[Math.floor(Math.random() * eventAttendants.length)]
+          .user;
+      const content = txtgen.paragraph(Math.ceil(Math.random() * 3));
```
AlbertoPdRF, line 81: As we always use `Math.floor()`, should we use it here for consistency too?

api/db/populate-comments.js
```diff
@@ -0,0 +1,94 @@
...
+      const parentComment =
```
AlbertoPdRF, line 82: A few issues with this:
- Only top-level comments can be parent comments of other comments, so we should enforce this here somehow
- What happens when `Math.random() <= 0.2` is `true` but `eventComments` is still empty? We should probably skip this altogether if `eventComments` is empty
- Should we make it more probable for a comment to have a parent comment? I guess most real conversation would occur in comment threads, no?
- If a comment doesn't have a parent comment, the `parentComment` value should not be set instead of being set to `null`. This can be easily fixed by only including the value on the `commentParams` object if it's not `null`

api/db/populate-comments.js
```diff
@@ -0,0 +1,94 @@
...
+      const commentParams = {
+        event,
+        user,
+        content,
+        parentComment,
+      };
+
+      const comment = await createComment(commentParams);
+      eventComments.push(comment);
+      commentsArray.push(comment);
```
AlbertoPdRF, line 94: Maybe we could leave this array out of the loop and just merge both arrays afterwards? This is not critical though, feel free to go either way

api/db/populate.js
```diff
@@ -48,9 +50,12 @@ const populateCollection = async () => {
     case 'friendships':
```
AlbertoPdRF: I just saw there's a missing break above this line.

api/db/populate.js
```diff
@@ -48,9 +50,12 @@ const populateCollection = async () => {
...
       await createFriendships(userArgs.v);
       break;
+    case 'comments':
+      await createComments(userArgs.v);
+      break;
     default:
       console.log(
-        `Unknown collection '${userArgs.c}', possible options are: [interests, users, events, attendants]`
+        `Unknown collection '${userArgs.c}', possible options are: [interests, users, events, attendants, friendships, comments]`
```
AlbertoPdRF: This seems like a rebase issue. While we are at it, I think it would be better to have `comments` before `friendships`, as it is already the case when populating the database, elsewhere -- I think this is just a bit of re-ordering in a few places.

AlbertoPdRF (changes requested):
Thanks for another nice PR submission @Be-Like!! 🚀

A few comments from me below 🙂

api/db/populate-comments.js
Be-Like: Well, if we did that then we would have to add 1, which would kind of defeat the purpose of doing the floor. Otherwise it would create a 0 sentence comment which I assume isn't allowed.

api/db/populate-comments.js
Be-Like: 2. If the event comments is empty, then parent comment will be null, which means it will just be a 'root' comment. This is just to simulate the feel for having replies to comments. 
3. What would you like it set to? 
4. I can do this, but in the end it is more work (miniscule though it may be) for the same result.

api/db/populate-comments.js
AlbertoPdRF: That's right -- maybe we should change it on the other places where we use are using the `Math.floor() + 1` approach instead!

api/db/populate-comments.js
AlbertoPdRF: 2. I see that in this case the `parentComment` is `undefined`, while in the other case is `null`. This is related to point 4, so I'll comment more below
3. Maybe `>= 0.5`? What do you think?
4. I think that the problem with this as is is that pushing a `null` value makes the document invalid, as it's not as defined on the comment schema and it wouldn't pass the validations that we have in place

api/db/drop.js
```diff
@@ -9,12 +9,14 @@ const User = require('../models/User');
 const Event = require('../models/Event');
 const Attendant = require('../models/Attendant');
 const Friendship = require('../models/Friendship');
+const Comment = require('../models/Comment');
```
AlbertoPdRF, line 11: I believe this should be moved one line above

api/db/drop.js
```diff
@@ -9,12 +9,14 @@ const User = require('../models/User');
...
 
 const getInterestsCount = async () => await Interest.countDocuments();
 const getUsersCount = async () => await User.countDocuments();
 const getEventsCount = async () => await Event.countDocuments();
 const getAttendantsCount = async () => await Attendant.countDocuments();
 const getFriendshipsCount = async () => await Friendship.countDocuments();
+const getCommentsCount = async () => await Comment.countDocuments();
```
AlbertoPdRF, line 18: Same here: I believe this should be moved one line above

api/db/drop.js
```diff
@@ -107,8 +117,19 @@ const dropFriendships = async (confirmed = false) => {
...
+const dropComments = async (confirmed = false) => {
+  console.log('\n', '\x1b[0m', 'Dropping comments collection...');
+  if (confirmed || (await getCommentsCount()) !== 0) {
+    await Comment.collection.drop();
+    console.log('\x1b[31m', 'Dropped comments collection');
+  } else {
+    console.log('\x1b[33m', 'Comments collection is already empty');
+  }
+};
+
```
AlbertoPdRF, line 119: And same here: I think this should go above the `dropFriendships` function definition

api/db/populate-events.js
```diff
@@ -76,7 +76,7 @@ const createEvents = async (multiplier, randomLocation, verbose) => {
       separator: ' ',
     });
 
-    const description = txtgen.article(Math.floor(Math.random() * 2) + 1);
+    const description = txtgen.article(Math.ceil(Math.random() * 2));
```
AlbertoPdRF: While we are at it, we could change this `2` for a `3`, as this was another suggestion of mine from when I thought that `Math.random()` could return `1`

AlbertoPdRF (commented):
These are 4 super small comments (and 3 of them are probably overly nitpicky), so if you don't want to bother addressing them don't worry, I'll merge the PR anyway when I'm back from having dinner! 🍴

AlbertoPdRF (approved):
Superb job @Be-Like, as usual! :shipit:

PR #344, opened 2020-10-12
Add scripts to populate and drop the messages collection

Be-Like:
Refer #279.

### DIFF
--- a/api/db/drop.js
+++ b/api/db/drop.js
@@ -141,7 +141,7 @@ const dropMessages = async (confirmed = false) => {
   console.log('\n', '\x1b[0m', 'Dropping messages collection...');
   if (confirmed || (await getMessagesCount()) !== 0) {
     await Message.collection.drop();
-    console.log('\x1b[33m', 'Dropped messages collection');
+    console.log('\x1b[31m', 'Dropped messages collection');
   } else {
     console.log('\x1b[33m', 'Messages collection is already empty');
   }
--- a/api/db/drop.js
+++ b/api/db/drop.js
@@ -141,9 +141,9 @@ const dropMessages = async (confirmed = false) => {
   console.log('\n', '\x1b[0m', 'Dropping messages collection...');
   if (confirmed || (await getMessagesCount()) !== 0) {
     await Message.collection.drop();
-    console.log('\x1b[31m', 'Dropped messages collection');
+    console.log('\x1b[33m', 'Dropped messages collection');
   } else {
-    console.log('\x1b[31m', 'Messages collection is already empty');
+    console.log('\x1b[33m', 'Messages collection is already empty');
   }
 };
 
--- a/api/db/drop.js
+++ b/api/db/drop.js
@@ -127,11 +127,19 @@ const dropFriendships = async (confirmed = false) => {
   } else {
     console.log('\x1b[33m', 'Friendships collection is already empty');
   }
+
+  if ((await getMessagesCount()) !== 0) {
+    const drop = await createPrompt(
+      'The messages collection is dependent on the friendships collection. Drop messages collection?'
+    );
+
+    if (drop) await dropMessages(drop);
+  }
 };
 
-const dropMessages = async () => {
+const dropMessages = async (confirmed = false) => {
   console.log('\n', '\x1b[0m', 'Dropping messages collection...');
-  if ((await getMessagesCount()) !== 0) {
+  if (confirmed || (await getMessagesCount()) !== 0) {
     await Message.collection.drop();
     console.log('\x1b[31m', 'Dropped messages collection');
   } else {
--- a/api/db/drop.js
+++ b/api/db/drop.js
@@ -10,6 +10,7 @@ const Event = require('../models/Event');
 const Attendant = require('../models/Attendant');
 const Comment = require('../models/Comment');
 const Friendship = require('../models/Friendship');
+const Message = require('../models/Message');
 
 const getInterestsCount = async () => await Interest.countDocuments();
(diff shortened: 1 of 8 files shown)


api/db/populate-events.js
```diff
@@ -12,12 +12,15 @@ const {
   getCountry,
   getRegion,
   getRandomSubset,
+  getArticle,
```
AlbertoPdRF: I believe that with this change the first line is not needed anymore, right?

api/db/populate-messages.js
```diff
@@ -0,0 +1,55 @@
...
+
+let displayLogs;
+
+const createMessage = async (messageParams) => {
+  const message = new Message({
+    friendship: messageParams.friendship,
+    user: messageParams.user,
+    content: messageParams.content,
+  });
+  message.save();
```
AlbertoPdRF: There's a missing `await` here

api/db/populate-messages.js
```diff
@@ -0,0 +1,55 @@
...
+
+const createMessages = async (verbose) => {
+  console.log('\n', '\x1b[0m', 'Populating messages collection...');
+  displayLogs = verbose;
+
+  const friendshipsArray = await Friendship.find();
+  const messagesArray = [];
+
+  for (const friendship of friendshipsArray) {
+    const messagesCount = Math.floor(Math.random() * 10);
```
AlbertoPdRF: I would say we should make this number bigger and instead make the content of every message smaller, as that could be a better model a real conversation? I'll leave a comment about this where the content of the message is generated too

api/db/populate-messages.js
```diff
@@ -0,0 +1,55 @@
...
+
+    for (let i = 0; i < messagesCount; i++) {
+      const user =
+        Math.random() < 0.5 ? friendship.requestant : friendship.receivant;
+      const content = getArticle();
```
AlbertoPdRF: As said above, I think a real conversation would consist mostly of smaller messages, so maybe we can throw a random number here to decide if we call the `getArticle()` function (should be improbable) or generate just a sentence. How does that sound?

api/db/drop.js
```diff
@@ -125,12 +127,31 @@ const dropFriendships = async (confirmed = false) => {
...
+  }
+};
+
+const dropMessages = async (confirmed = false) => {
+  console.log('\n', '\x1b[0m', 'Dropping messages collection...');
+  if (confirmed || (await getMessagesCount()) !== 0) {
+    await Message.collection.drop();
+    console.log('\x1b[31m', 'Dropped messages collection');
+  } else {
+    console.log('\x1b[31m', 'Messages collection is already empty');
```
AlbertoPdRF: The first param should be `'\x1b[33m'`

AlbertoPdRF (changes requested):
This looks mostly good @Be-Like, I've just left a few comments inline. Thanks for your work on another great PR! :tada:

api/db/populate-messages.js
Be-Like: That's fine. However, I would suggest we reconsider the quantity of users and/or events we are creating. It is beginning to take a very long time to generate all the collections. And most of these items we only really need a handful of records to get a clear idea of how the application will behave.

AlbertoPdRF: Yeah, I think we can reduce those by a factor of ten, and if someone needs more they can just use the `-m` argument for that

api/db/drop.js
```diff
@@ -141,9 +141,9 @@ const dropMessages = async (confirmed = false) => {
   console.log('\n', '\x1b[0m', 'Dropping messages collection...');
   if (confirmed || (await getMessagesCount()) !== 0) {
     await Message.collection.drop();
-    console.log('\x1b[31m', 'Dropped messages collection');
+    console.log('\x1b[33m', 'Dropped messages collection');
```
AlbertoPdRF: This one was correct as `'\x1b[31m'`, so let's change it back

api/db/populate-events.js
```diff
@@ -71,15 +70,15 @@ const createEvents = async (multiplier, randomLocation, verbose) => {
   const now = new Date();
   const futureDate = new Date().setFullYear(now.getFullYear() + 5);
 
-  for (let i = 0; i < 500 * multiplier; i++) {
+  for (let i = 0; i < 50 * multiplier; i++) {
```
AlbertoPdRF, line 73: Please note that this change (and the users' one) requires a small update on the API's README.

api/db/populate-messages.js
```diff
@@ -27,12 +27,12 @@ const createMessages = async (verbose) => {
...
 
   for (const friendship of friendshipsArray) {
-    const messagesCount = Math.floor(Math.random() * 10);
+    const messagesCount = Math.floor(Math.random() * 25);
 
     for (let i = 0; i < messagesCount; i++) {
       const user =
         Math.random() < 0.5 ? friendship.requestant : friendship.receivant;
-      const content = getArticle();
+      const content = getParagraph();
```
AlbertoPdRF, line 35: Maybe we could pass a `Math.ceil(Math.random() * 3)` parameter to this `getParagraph()` function?

AlbertoPdRF (changes requested):
Just 3 more small comments to address and we are good to go! :ship:

api/db/populate-messages.js
AlbertoPdRF: Oops, this comment doesn't make sense, please disregard it

AlbertoPdRF (approved):
This LGTM, thanks @Be-Like! :100:

PR #347, opened 2020-10-13
Populate (and drop) the notifications collection

### LINKED ISSUE #279: Populate the development database with dummy data

Be-Like:
Closes #279.

You might have a lot to say about this one... 😅 I tried to clean it up as best I could, but it is a pretty complex script. 

Also, as discussed in Slack, it might be worth considering not including this in the standard `npm run db:populate` command as this is a pretty pricey script to run. And it is something that is already slightly tested just when signing up.

Also, I didn't include the first four types in the Notifications enum because it is easier to test that by just signing up a new user. And most of those have already been done just by virtue of running the populate script.

### DIFF
--- a/api/db/drop.js
+++ b/api/db/drop.js
@@ -163,14 +163,14 @@ const dropNotifications = async (confirmed = false) => {
     await Notification.collection.drop();
     console.log('\x1b[31m', 'Dropped notifications collection');
   } else {
-    console.log('\x1b[31m', 'Notifications collection is already empty');
+    console.log('\x1b[33m', 'Notifications collection is already empty');
   }
 };
 
 const dropCollections = async () => {
   connectDb();
-  await dropMessages();
   await dropNotifications();
+  await dropMessages();
   await dropFriendships();
   await dropComments();
   await dropAttendants();
--- a/api/db/drop.js
+++ b/api/db/drop.js
@@ -11,6 +11,7 @@ const Attendant = require('../models/Attendant');
 const Comment = require('../models/Comment');
 const Friendship = require('../models/Friendship');
 const Message = require('../models/Message');
+const Notification = require('../models/Notification');
 
 const getInterestsCount = async () => await Interest.countDocuments();
 const getUsersCount = async () => await User.countDocuments();
@@ -19,6 +20,7 @@ const getAttendantsCount = async () => await Attendant.countDocuments();
 const getCommentsCount = async () => await Comment.countDocuments();
 const getFriendshipsCount = async () => await Friendship.countDocuments();
 const getMessagesCount = async () => await Message.countDocuments();
+const getNotificationsCount = async () => await Notification.countDocuments();
 
 const userArgs = minimist(process.argv.slice(2), {
   string: 'collection',
@@ -69,6 +71,14 @@ const dropUsers = async (confirmed = false) => {
     );
 
     if (drop) await dropFriendships(drop);
+  }
+
+  if ((await getNotificationsCount()) !== 0) {
+    const drop = await createPrompt(
+      'The notifications collection is dependent on the users collection. Drop notifications collection?'
+    );
+
+    if (drop) await dropNotifications(drop);
   }
 };
 
@@ -147,9 +157,20 @@ const dropMessages = async (confirmed = false) => {
   }
 };
 
+const dropNotifications = async (confirmed = false) => {
+  console.log('\n', '\x1b[0m', 'Dropping notifications collection');
+  if (confirmed || (await getNotificationsCount()) !== 0) {
+    await Notification.collection.drop();
+    console.log('\x1b[31m', 'Dropped notifications collection');
+  } else {
+    console.log('\x1b[31m', 'Notifications collection is already empty');
+  }
+};
+
 const dropCollections = async () => {
   connectDb();
   await dropMessages();
+  await dropNotifications();
   await dropFriendships();
   await dropComments();
   await dropAttendants();
@@ -183,9 +204,12 @@ const dropCollection = async () => {
     case 'messages':
       await dropMessages();
       break;
+    case 'notifications':
+      await dropNotifications();
+      break;
     default:
       console.log(
-        `Unknown collection '${userArgs.c}', possible options are: [interests, users, events, attendants, comments, friendships, messages]`
+        `Unknown collection '${userArgs.c}', possible options are: [interests, users, events, attendants, comments, friendships, messages, notifications]`
       );
       break;
   }
--- a/api/db/populate-attendants.js
+++ b/api/db/populate-attendants.js
@@ -15,13 +15,16 @@ const createAttendant = async (attendantParams) => {
   await attendant.save();
 
   // New attendee notification
-  await createNotification({
-    user: attendantParams.event.createdBy,
-    type: notificationTypes[4],
-    read: false,
-    referencedUser: attendant.user,
-    referencedEvent: attendant.event,
-  });
+  await createNotification(
+    {
+      user: attendantParams.event.createdBy,
+      type: notificationTypes[4],
+      read: false,
+      referencedUser: attendant.user,
+      referencedEvent: attendant.event,
+    },
+    displayLogs
+  );
 
   if (displayLogs) {
     console.log('\n', '\x1b[0m', `New attendant created: ${attendant}`);
--- a/api/db/populate-attendants.js
+++ b/api/db/populate-attendants.js
@@ -1,4 +1,5 @@
 const { getRandomSubset, createPrompt } = require('./utils');
+const { createNotification } = require('./populate-notifications');
 
 const User = require('../models/User');
 const Event = require('../models/Event');
@@ -7,11 +8,17 @@ const Attendant = require('../models/Attendant');
 let displayLogs;
 
 const createAttendant = async (attendantParams) => {
-  const attendant = new Attendant({
-    event: attendantParams.event,
-    user: attendantParams.user,
-  });
+  const attendant = new Attendant(attendantParams);
   await attendant.save();
+
+  // New attendee notification
+  await createNotification({
+    user: attendantParams.event.createdBy,
+    type: 'newAttendant',
+    read: false,
+    referencedUser: attendant.user,
+    referencedEvent: attendant.event,
+  });
 
   if (displayLogs) {
     console.log('\n', '\x1b[0m', `New attendant created: ${attendant}`);
--- a/api/db/populate-attendants.js
+++ b/api/db/populate-attendants.js
@@ -9,22 +9,26 @@ const Event = require('../models/Event');
 const Attendant = require('../models/Attendant');
 
 let displayLogs;
+let notificationsCount = 0;
 
 const createAttendant = async (attendantParams) => {
   const attendant = new Attendant(attendantParams);
   await attendant.save();
 
   // New attendee notification
-  await createNotification(
-    {
-      user: attendantParams.event.createdBy,
-      type: notificationTypes[4],
-      read: false,
-      referencedUser: attendant.user,
-      referencedEvent: attendant.event,
-    },
-    displayLogs
-  );
+  if (attendant.user.toString() !== attendant.event.createdBy.toString()) {
+    await createNotification(
+      {
+        user: attendant.event.createdBy,
+        type: notificationTypes[4],
+        read: false,
+        referencedUser: attendant.user,
+        referencedEvent: attendant.event,
+      },
+      displayLogs
+    );
+    notificationsCount++;
+  }
 
   if (displayLogs) {
     console.log('\n', '\x1b[0m', `New attendant created: ${attendant}`);
@@ -100,6 +104,10 @@ const createAttendants = async (verbose) => {
   }
 
   console.log('\x1b[32m', `Created ${attendantsArray.length} attendants`);
+  console.log(
+    '\x1b[32m',
+    `Created ${notificationsCount} attendant notifications`
+  );
 };
 
 module.exports = { createAttendants };
--- a/api/db/populate-attendants.js
+++ b/api/db/populate-attendants.js
@@ -1,5 +1,8 @@
 const { getRandomSubset, createPrompt } = require('./utils');
-const { createNotification } = require('./populate-notifications');
+const {
+  createNotification,
+  notificationTypes,
+} = require('./populate-notifications');
 
 const User = require('../models/User');
 const Event = require('../models/Event');
@@ -14,7 +17,7 @@ const createAttendant = async (attendantParams) => {
   // New attendee notification
   await createNotification({
     user: attendantParams.event.createdBy,
-    type: 'newAttendant',
+    type: notificationTypes[4],
     read: false,
     referencedUser: attendant.user,
     referencedEvent: attendant.event,
--- a/api/db/populate-attendants.js
+++ b/api/db/populate-attendants.js
@@ -103,10 +103,9 @@ const createAttendants = async (verbose) => {
     }
   }
 
-  console.log('\x1b[32m', `Created ${attendantsArray.length} attendants`);
   console.log(
     '\x1b[32m',
-    `Created ${notificationsCount} attendant notifications`
+    `Created ${attendantsArray.length} attendants (and ${notificationsCount} related notifications)`
   );
 };
 
--- a/api/db/populate-comments.js
+++ b/api/db/populate-comments.js
@@ -35,10 +35,7 @@ const prerequisites = (eventsCount, attendantsCount) => {
 };
 
 const getParentComment = (comments) => {
-  const parentComments = comments.filter(
-    (comment) =>
-      comment.parentComment === undefined || comment.parentComment === null
-  );
+  const parentComments = comments.filter((comment) => !comment.parentComment);
   return parentComments[Math.floor(Math.random() * parentComments.length)];
 };
 
@@ -109,10 +106,9 @@ const createComments = async (verbose) => {
     }
   }
 
-  console.log('\x1b[32m', `Created ${commentsArray.length} comments`);
   console.log(
     '\x1b[32m',
-    `Created ${notificationsCount} comment notifications`
+    `Created ${commentsArray.length} comments (and ${notificationsCount} related notifications)`
   );
 };
 
--- a/api/db/populate-comments.js
+++ b/api/db/populate-comments.js
@@ -3,6 +3,7 @@ const { getParagraph } = require('./utils');
 const Event = require('../models/Event');
 const Attendant = require('../models/Attendant');
 const Comment = require('../models/Comment');
+const { createNotification } = require('./populate-notifications');
 
 let displayLogs;
 
@@ -37,11 +38,7 @@ const getParentComment = (comments) => {
 };
 
 const createComment = async (commentParams) => {
-  const comment = new Comment({
-    event: commentParams.event,
-    user: commentParams.user,
-    content: commentParams.content,
-  });
+  const comment = new Comment(commentParams);
   if (commentParams.parentComment) {
     comment.parentComment = commentParams.parentComment;
   }
@@ -92,6 +89,17 @@ const createComments = async (verbose) => {
       const comment = await createComment(commentParams);
       eventComments.push(comment);
       commentsArray.push(comment);
+
+      for (const attendee of eventAttendants) {
+        if (attendee.user.toString() === comment.user.toString()) continue;
+        await createNotification({
+          user: attendee.user,
+          type: 'newComment',
+          read: true,
+          referencedUser: comment.user,
+          referencedEvent: comment.event,
+        });
+      }
     }
   }
 
--- a/api/db/populate-comments.js
+++ b/api/db/populate-comments.js
@@ -82,8 +82,8 @@ const createComments = async (verbose) => {
         event,
         user,
         content,
-        parentComment,
       };
+      if (parentComment) commentParams.parentComment = parentComment;
 
       const comment = await createComment(commentParams);
       eventComments.push(comment);
--- a/api/db/populate-comments.js
+++ b/api/db/populate-comments.js
@@ -95,13 +95,16 @@ const createComments = async (verbose) => {
 
       for (const attendee of eventAttendants) {
         if (attendee.user.toString() === comment.user.toString()) continue;
-        await createNotification({
-          user: attendee.user,
-          type: notificationTypes[5],
-          read: true,
-          referencedUser: comment.user,
-          referencedEvent: comment.event,
-        });
+        await createNotification(
+          {
+            user: attendee.user,
+            type: notificationTypes[5],
+            read: true,
+            referencedUser: comment.user,
+            referencedEvent: comment.event,
+          },
+          displayLogs
+        );
       }
     }
   }
--- a/api/db/populate-comments.js
+++ b/api/db/populate-comments.js
@@ -1,14 +1,15 @@
 const { getParagraph } = require('./utils');
-
-const Event = require('../models/Event');
-const Attendant = require('../models/Attendant');
-const Comment = require('../models/Comment');
 const {
   createNotification,
   notificationTypes,
 } = require('./populate-notifications');
 
+const Event = require('../models/Event');
+const Attendant = require('../models/Attendant');
+const Comment = require('../models/Comment');
+
 let displayLogs;
+let notificationsCount = 0;
 
 const prerequisites = (eventsCount, attendantsCount) => {
   if (eventsCount === 0 && attendantsCount === 0) {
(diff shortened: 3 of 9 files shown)


api/db/drop.js
```diff
@@ -147,9 +157,20 @@ const dropMessages = async (confirmed = false) => {
   }
 };
 
+const dropNotifications = async (confirmed = false) => {
+  console.log('\n', '\x1b[0m', 'Dropping notifications collection');
+  if (confirmed || (await getNotificationsCount()) !== 0) {
+    await Notification.collection.drop();
+    console.log('\x1b[31m', 'Dropped notifications collection');
+  } else {
+    console.log('\x1b[31m', 'Notifications collection is already empty');
```
AlbertoPdRF:
```suggestion
    console.log('\x1b[33m', 'Notifications collection is already empty');
```

api/db/drop.js
```diff
@@ -147,9 +157,20 @@ const dropMessages = async (confirmed = false) => {
...
+  }
+};
+
 const dropCollections = async () => {
   connectDb();
   await dropMessages();
+  await dropNotifications();
```
AlbertoPdRF:
```suggestion
  await dropNotifications();
  await dropMessages();
```

AlbertoPdRF:
First of all, thanks @Be-Like for working on these awesome population scripts! :100:

I'm leaving a high level comment instead of a full code review so we can discuss about the approach followed in this one: I see that you define a `notificationsCount` and, from there, for each user, generate a random number of random notifications until that `notificationsCount` is met, did I get that right? What I thought we would be doing instead is one of:
- Get an array of all the things that have been created that, on the real application, would generate a notification, loop through this array, and generate all notifications
- On each population script, when something that, on the real application, would generate a notification, gets created, generate the corresponding notification

Did you have any of those approaches in mind before proceeding with yours? What made me think of them is that, as the script is right now, we would be getting a notifications collection pretty different from the real one. Also, regarding your comments, if we see that generating all notifications is very pricey, we can for sure add an argument that let's the developer opt out (or in) of it.

Be-Like:
So I just generated a handful of notifications for each user. Those notifications are legit and relate to that users friendship/event/etc. I think it would be impractical to generate all those notifications for each user, when we as the developers won't be logging in to most, or any, of the users generated.

AlbertoPdRF:
I think that we could expect for developers to at least use the admin user, so generating all notifications for that one sounds reasonable to me. Generating the notifications for the others could be handled with an opt out (or in) argument, especially if it takes too long. How does that sound to you?

Be-Like:
Would you like all the different notification types (confirmEmail, createEvent, uploadAvatar, and selectInterests) as well?

AlbertoPdRF:
Yep, I would add all types.

AlbertoPdRF:
I just ran the script and the result looks amazing @Be-Like! :heart_eyes:

I'll provide a full code review later today :slightly_smiling_face:

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+  if (notification.read) notification.readAt = new Date();
+  if (notificationParams.referencedUser) {
+    notification.referencedUser = notificationParams.referencedUser;
+  }
+  if (notificationParams.referencedEvent) {
+    notification.referencedEvent = notificationParams.referencedEvent;
+  }
+  if (notificationParams.referencedFriendship) {
+    notification.referencedFriendship = notificationParams.referencedFriendship;
+  }
```
AlbertoPdRF:
```suggestion
  const notification = new Notification(notificationParams);
  if (notification.read) notification.readAt = new Date();
```
Somehow I overlooked this simplification in all previous code reviews... :facepalm:

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+    notificationsArray.push(
+      await createNotification({
+        user,
+        type: types[0],
+        read,
+      })
+    );
+    // Create event notification
+    read = eventsList.some(
+      (event) => event.createdBy.toString === user._id.toString()
```
AlbertoPdRF:
```suggestion
      (event) => event.createdBy.toString() === user._id.toString()
```

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+    );
+    notificationsArray.push(
+      await createNotification({
+        user,
+        type: types[1],
+        read,
+      })
+    );
+    // Create upload avatar notification
```
AlbertoPdRF:
```suggestion
    // Upload avatar notification
```

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+    read = user.avatar ? true : false;
+    notificationsArray.push(
+      await createNotification({
+        user,
+        type: types[2],
+        read,
+      })
+    );
+    // Selected interests notification
```
AlbertoPdRF:
```suggestion
    // Select interests notification
```

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+    const interests = await Interest.find({ user });
+    read = interests.length > 0;
```
AlbertoPdRF:
```suggestion
    read = user.interests.length > 0;
```
I think the the `read` as it was defined would never be `true`, as interests don't have any `user` value

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+  const eventsList = await Event.find();
+  const attendantsList = await Attendant.find();
+  const commentsList = await Comment.find();
+  const friendshipsList = await Friendship.find();
+  const messagesList = await Message.find();
+
+  // For each user
+  for (const user of usersList) {
+    let read = user.emailConfirmed;
+    // Email confirmation notification
```
AlbertoPdRF:
```suggestion
    // Confirm email notification
```

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+    read = interests.length > 0;
+    notificationsArray.push(
+      await createNotification({
+        user,
+        type: types[3],
+        read,
+      })
+    );
+
+    // Create attendee and comment notifications
```
AlbertoPdRF:
```suggestion
    // New attendee and new comment notifications
```

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+    const usersEvents = eventsList.filter(
+      (event) => event.createdBy.toString() === user._id.toString()
+    );
```
AlbertoPdRF: As above we are using an `eventsList.some(...)` with the same condition, maybe it would make sense to move this `eventsList.filter(...)` above and, to get the needed boolean value, just check if the `usersEvents.length > 0` instead of using the `some`.

Also: should be `userEvents`?

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+          referencedEvent: event,
+        };
+        notificationsArray.push(await createNotification(notificationParams));
+      }
+
+      const comments = commentsList.filter(
+        (comment) =>
+          comment.event.toString() === event._id.toString() &&
+          comment.user.toString() !== user._id.toString()
+      );
```
AlbertoPdRF: This will only generate the new comment notification for the creator of the event, no? It should be generated for all the event attendees to replicate the real application behavior

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+          type: types[5],
+          read,
+          referencedUser: comment.user,
+          referencedEvent: comment.event,
+        };
+        notificationsArray.push(await createNotification(notificationParams));
+      }
+    }
+
+    // New friendship request notification
```
AlbertoPdRF:
```suggestion
    // New friendship request notifications
```

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+    const friendRequests = friendshipsList.filter(
```
AlbertoPdRF: Nitpick: should be `friendshipRequests`, no?

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+        type: types[6],
+        read,
+        referencedUser: friendRequest.requestant,
+        referencedFriendship: friendRequest,
+      };
+      notificationsArray.push(await createNotification(notificationParams));
+    }
+
+    // Accepted friendship request notifications
+    const acceptedFriendships = friendshipsList.filter(
```
AlbertoPdRF: Nitpick: for consistency, I think it should be `acceptedFriendshipRequests`

api/db/populate-notifications.js
```diff
@@ -0,0 +1,222 @@
...
+        type: types[7],
+        read,
+        referencedUser: acceptedFriendship.receivant,
+        referencedFriendship: acceptedFriendship,
+      };
+      notificationsArray.push(await createNotification(notificationParams));
+    }
+
+    // New message notifications
+    const userfriendships = friendshipsList.filter(
```
AlbertoPdRF: Nitpick: should be `userFriendships`

AlbertoPdRF (changes requested):
This looks mostly good @Be-Like! :top: I've just left a few comments inline, hopefully easy to resolve!

I have a few more general ones:
1. Do you think it would make sense to, instead of having a separate script for creating all the notifications, generate each notification when the document that would trigger its creation gets created? Most of the already existing code could be re-used for this, and I'm asking it because doing that would fix a few problems:
   - The notifications collection wouldn't need to be dropped and re-generated every time the population script is ran with a non-empty database, which can become very costly as the other collections grow
   - All notifications would have a `createdAt` value more consistent with the `createdAt` value of the document that triggered the notification creation
   - If the developer populated any collections individually, their respective notifications would get created too
2. There are some of the `read` values that aren't incompatible with what would happen on the real app. Let me illustrate that with an example: if there are two notifications about two comments of the same event, and the newest one is marked as read, it's impossible to have the first one marked as unread on the real application, as visiting the comments page of the event (i.e. "reading" the second notification) would mark both as read.
3. Do you still plan to send a follow up PR addressing the details we've left behind all the other PRs (add users' avatars, remove the time part of the users' birthdate datetime value, add events' cover photos, ...)? If that's the case, maybe we should leave the `Closes #279` message for that PR, or we can open a new issue for that too if you prefer!

And, of course, thanks again for your awesome work! :sparkles:

Be-Like:
1. I thought about this as well and there was a reason I didn't go that route (other than I was already halfway through 😅) that I can no longer recall. The only thing I can think of right now is that if the user is testing anything related to notifications, they would have to rerun the other scripts (friendships for example). If they do that so many times, eventually there would be no more documents generated (because everybody is already friends). Also, it would be harder to know for whom notifications were generated. As it is now, at the very least we know that Admin will have all the notifications generated each run and we won't have to search the new notifications to see which users they relate to. In either case, I imagine this would be somewhat easy to do, just more of a hassle.
2. I thought about this, and can look into it a little more. Unless the collection query returns all the documents in created by order (oldest to newest), it will add some more complexity and time to the population script. As it is, it generates about 900 documents for every 10 users. 
3. We can make those changes in #279. It doesn't matter either way to me. 

I'll begin working on the comments you left above.

AlbertoPdRF:
1. What you say makes sense, although I guess that if someone wanted to test something related to notifications that they couldn't already test with the existing ones, they would need brand new ones instead of re-generating the old ones. Thus, I'm inclined to think that the other solution would solve more problems than it would create, but let's keep discussing this if you are not convinced or if other reasons to keep the current approach come to your mind!
2. Getting the collection sorted by `createdAt` can be easily accomplished, what I don't know is the price we'll have to pay to handle this properly. I guess we can look into it to have an idea and then decide on how to proceed.
3. Feel free to go either way! 🙂

api/db/populate-attendants.js
```diff
@@ -7,12 +11,21 @@ const Attendant = require('../models/Attendant');
...
-    event: attendantParams.event,
-    user: attendantParams.user,
-  });
+  const attendant = new Attendant(attendantParams);
   await attendant.save();
 
+  // New attendee notification
+  await createNotification(
+    {
+      user: attendantParams.event.createdBy,
```
AlbertoPdRF: I guess `attendant.event.createdBy` would not work here?

api/db/populate-comments.js
```diff
@@ -3,6 +3,10 @@ const { getParagraph } = require('./utils');
 const Event = require('../models/Event');
 const Attendant = require('../models/Attendant');
 const Comment = require('../models/Comment');
+const {
+  createNotification,
+  notificationTypes,
+} = require('./populate-notifications');
```
AlbertoPdRF, line 5: Let's move this to line 2, and leave a blank line after it.

api/db/populate-comments.js
```diff
@@ -37,11 +41,7 @@ const getParentComment = (comments) => {
...
 const createComment = async (commentParams) => {
-  const comment = new Comment({
-    event: commentParams.event,
-    user: commentParams.user,
-    content: commentParams.content,
-  });
+  const comment = new Comment(commentParams);
   if (commentParams.parentComment) {
     comment.parentComment = commentParams.parentComment;
   }
```
AlbertoPdRF: I believe this is not necessary anymore with the change above!

api/db/populate-events.js
```diff
@@ -105,6 +96,16 @@ const createEvents = async (multiplier, randomLocation, verbose) => {
     eventsArray.push(await createEvent(eventParams));
   }
 
+  const uniqueEventCreators = [
+    ...new Set(eventsArray.map((event) => event.createdBy)),
+  ];
+  for (const creator of uniqueEventCreators) {
```
AlbertoPdRF: Maybe rename `creator` to `eventCreator`?

api/db/populate-friendships.js
```diff
@@ -1,19 +1,42 @@
...
+    {
+      user: friendship.receivant,
+      type: notificationTypes[6],
+      read: friendship.accepted,
+      referencedUser: friendship.requestant,
+      referencedFriendship: friendship,
+    },
+    displayLogs
+  );
+  if (friendship.accepted) {
```
AlbertoPdRF, line 30: I would add a comment before this line identifying the type of the notification too, as you did for all others.

api/db/populate-messages.js
```diff
@@ -13,6 +17,21 @@ const createMessage = async (messageParams) => {
   });
```
AlbertoPdRF: This could be changed too so only the full `messageParams` is passed to the `message` constructor.

api/db/populate-messages.js
```diff
@@ -13,6 +17,21 @@ const createMessage = async (messageParams) => {
...
   await message.save();
 
+  const user =
+    message.friendship.receivant.toString() === message.user.toString()
+      ? message.friendship.receivant
+      : message.friendship.requestant;
```
AlbertoPdRF: This should be the other way around.

Also, maybe it would help make it a bit more understandable to write `message.user.toString() === message.friendship.receivant.toString()` instead of as it's written now, but feel free to keep it as is!

api/db/populate-notifications.js
```diff
@@ -0,0 +1,27 @@
...
+  'newComment',
+  'newFriendshipRequest',
+  'acceptedFriendshipRequest',
+  'newMessage',
+];
+
+const createNotification = async (notificationParams, verbose = false) => {
+  const notification = new Notification(notificationParams);
+  if (notification.read) notification.readAt = new Date();
+
```
AlbertoPdRF: Nitpick: we are not leaving a blank line here on the other scripts

api/db/populate-notifications.js
```diff
@@ -0,0 +1,27 @@
...
+  await notification.save();
+
+  if (verbose) {
+    console.log('\n', '\x1b[0m', `New notification created: ${notification}`);
+  }
+  return notification;
+};
+
+module.exports = { createNotification, notificationTypes };
```
AlbertoPdRF: Nitpick: let's change the order of this

api/db/populate.js
```diff
@@ -61,7 +61,7 @@ const populateCollection = async () => {
       break;
     default:
       console.log(
-        `Unknown collection '${userArgs.c}', possible options are: [interests, users, events, attendants, comments, friendships, messages]`
+        `Unknown collection '${userArgs.c}', possible options are: [interests, users, events, attendants, comments, friendships, messages, notifications]`
```
AlbertoPdRF: I think this should go now, as the notifications collection can't be populated standalone

api/db/populate-comments.js
```diff
@@ -92,6 +92,20 @@ const createComments = async (verbose) => {
...
       eventComments.push(comment);
       commentsArray.push(comment);
+
+      for (const attendee of eventAttendants) {
+        if (attendee.user.toString() === comment.user.toString()) continue;
+        await createNotification(
+          {
+            user: attendee.user,
+            type: notificationTypes[5],
+            read: true,
```
AlbertoPdRF: Shouldn't these all be `false` by default, as it's done with most of the other collections?

api/db/populate-attendants.js
```diff
@@ -7,12 +11,21 @@ const Attendant = require('../models/Attendant');
...
 const createAttendant = async (attendantParams) => {
-  const attendant = new Attendant({
-    event: attendantParams.event,
-    user: attendantParams.user,
-  });
+  const attendant = new Attendant(attendantParams);
   await attendant.save();
 
+  // New attendee notification
+  await createNotification(
```
AlbertoPdRF: This notification has to be created in all cases but when the referenced user is the event creator

AlbertoPdRF (changes requested):
This looks great @Be-Like! :star_struck:

Apart from the comments inline, I think I would add somewhere (maybe on the output logged to the console? or on the API's README?) that the notifications collection has been populated with the corresponding notifications when that's the case. How does that sound?

Other than that, we are pretty close to shipping this one :ship: :eyes:

Two notes for the follow up PR:
- I haven't yet found an event with two or more parent comments, but I've found many with one parent comment that has many child comments. We should check if the logic is flawed or biased here, or maybe I just saw this pattern by random chance!
- The names of events are a bit weird as they are. I think we should change how the `dictionaries` for them are defined (i.e. `dictionaries: [adjectives, animals, colors, starWars]`) to be `dictionaries: [adjectives, colors, animals]`. As the `length` of the names is defined as `3`, the `starWars` dictionary is never used, and it would read better if the color was before the animal than after it. Also, if we get rid of the `starWars` dictionary, we no longer need to specify the `length` of the name!

api/db/populate-attendants.js
```diff
@@ -100,6 +104,10 @@ const createAttendants = async (verbose) => {
   }
 
   console.log('\x1b[32m', `Created ${attendantsArray.length} attendants`);
+  console.log(
+    '\x1b[32m',
+    `Created ${notificationsCount} attendant notifications`
+  );
```
AlbertoPdRF: Maybe, to avoid too much output, all of this could be merged in one line, something like:

```javascript
console.log(
  '\x1b[32m',
  `Created ${attendantsArray.length} attendants (and ${notificationsCount} related notifications)`
);

```
I don't feel strongly about this though, so feel free to leave it as is!

api/db/populate-comments.js
```diff
@@ -35,16 +36,14 @@ const prerequisites = (eventsCount, attendantsCount) => {
 
 const getParentComment = (comments) => {
   const parentComments = comments.filter(
-    (comment) => comment.parentComment === undefined
+    (comment) =>
+      comment.parentComment === undefined || comment.parentComment === null
```
AlbertoPdRF: Can't this be simply written as `(comment) => !comment.parentComment`? I see now where this error came from though, so I'll leave a comment about it right below.

api/db/populate-comments.js
```diff
@@ -35,16 +36,14 @@ const prerequisites = (eventsCount, attendantsCount) => {
...
   );
   return parentComments[Math.floor(Math.random() * parentComments.length)];
 };
 
 const createComment = async (commentParams) => {
   const comment = new Comment(commentParams);
```
AlbertoPdRF, line 43: This is where the error above came from: before this line, we should check if `commentParams` has a `parentComment` value of `null` and, if yes, remove it from the object before passing it to the comment constructor. Another possible solution, probably more clear, would be to, on the `createComments` function, only add the `parentComment` value to the `commentParams` object if it's not `null`. If you agree with it, let's go with this second approach, which by the way would be moving the few lines of code that we just deleted below this to the appropriate place and with the appropriate names to the `createComments` function

api/db/populate-users.js
```diff
@@ -139,6 +141,10 @@ const createUsers = async (multiplier, randomLocation, verbose) => {
   }
 
   console.log('\x1b[32m', `Created ${usersArray.length} regular users`);
+  console.log(
+    '\x1b[32m',
+    `Created ${notificationsCount} notifications related to creation of a new user`
```
AlbertoPdRF: It feels a bit weird to me that we include here the number of notifications related to the admin user but just above only count the regular users we have created :confused:

api/db/populate-messages.js
```diff
@@ -1,18 +1,35 @@
...
-    user: messageParams.user,
-    content: messageParams.content,
-  });
+  const message = new Message(messageParams);
   await message.save();
 
+  const user =
+    message.friendship.receivant.toString() === message.user.toString()
+      ? message.friendship.receivant
+      : message.friendship.requestant;
```
AlbertoPdRF:
```suggestion
      ? message.friendship.requestant
      : message.friendship.receivant;
```
This is what I mentioned earlier that should be the other way around, sorry if the comment was a bit confusing!

Also, I just realized that we are generating messages for all friendships, but only accepted friendships do have chats on the real application, so we need to address this below! A simple check should do :slightly_smiling_face:

AlbertoPdRF (changes requested):
We are almost there! :pinching_hand:

api/db/populate-comments.js
Be-Like: The latter is how it was, but we had decided to simplify it so we could so we could just pass the params to the newly created document (eg. `new Comment(commentParams)` rather than `new Comment({...})`). 

Personally, I believe it is better to include those fields and to leave it as null if it is irrelevant. It will make handling the object much easier as well as providing better structure for the documents (even if that means that the value of a field is null).

api/db/populate-users.js
Be-Like: Easiest solution would be to include the admin in the count. It's kind of irrelevant whether or not to count admin in the logging.

api/db/populate-comments.js
AlbertoPdRF: Yes, I know, that's why we now have the problem again, so the proposed solution is to handle it during the creation of the `commentParams` object instead of here. I would say that we should go that way for now and then we can decide if we want to allow documents to have null values

api/db/populate-users.js
AlbertoPdRF: Yep, let's just do that (removing the `regular` word of the sentence)!

api/db/populate-messages.js
```diff
@@ -16,19 +16,21 @@ const createMessage = async (messageParams) => {
...
-      read: false,
-      referencedUser: message.user,
-      referencedFriendship: message.friendship,
-    },
-    displayLogs
-  );
-  notificationsCount++;
+      ? message.friendship.requestant
+      : message.friendship.receivant;
+  if (!message.friendship.accepted) {
```
AlbertoPdRF: This check shouldn't go here but somewhere below -- what doesn't have to be created if the friendship request hasn't been accepted is the actual message instead of its notification

api/db/populate-users.js
```diff
@@ -140,10 +140,9 @@ const createUsers = async (multiplier, randomLocation, verbose) => {
     usersArray.push(await createUser(userParams));
   }
 
-  console.log('\x1b[32m', `Created ${usersArray.length} regular users`);
   console.log(
     '\x1b[32m',
-    `Created ${notificationsCount} notifications related to creation of a new user`
+    `Created ${usersArray.length} users (and ${notificationsCount} related notifications)`
```
AlbertoPdRF, line 147: Did we agree to count the admin here too in the end? This can be easily done by pushing it too to the users' array. And while we are at it, the function that creates the admin user is missing the `async` and `await` keywords

AlbertoPdRF (changes requested):

api/db/populate-users.js
```diff
@@ -132,7 +134,7 @@ const createUsers = async (multiplier, randomLocation, verbose) => {
       dateOfBirth: getRandomDate(
         new Date(1970, 0, 1),
         new Date().setFullYear(now.getFullYear() - 13)
-      ),
+      ).setUTCHours(0, 0, 0, 0),
```
AlbertoPdRF, line 137: This should also be applied to the admin user, although there you can write it directly as `new Date(2000, 0, 1, 0, 0, 0, 0)`

AlbertoPdRF (approved):
LGTM! GREAT work @Be-Like!! :tada:

PR #348, opened 2020-10-17
305 navcomp refactoring

### LINKED ISSUE #305: Refactor de `NavigationBarAndDrawer` component

The `NavigationBarAndDrawer` component has the navigation bar, a drawer, and two popovers, which makes it too big and thus difficult to read. Also, there's probably some code duplication that shouldn't be there.

I'd like for that component to be refactored into multiple, smaller ones, as we better see fit.

Leaving the huge component be.

**Additional context**

Making the component easier to read could help people looking to address #231.

san-oop:
Fixes #305 

The single big "NavigationBarAndDrawer" component is now refactored in to smaller sub components; It is mainly divided in to 3 more components

1. NavigationBar : The app bar
2. NavigationDrawer: The hamburger menu
3. NavigationPopover: Popovers for notifications which come in the app bar

**NavigationDrawer** is designed in such a way that an array of menuitems is looped to get each menu item. So if a new menu item is required to be added in future, it is as easy by just adding a new object to this array of menuitems. (Menuitems is further divided in to 2, primary and secondary: refer the code in NavigationDrawer)

**NavigationPopover** This is not a generic popover component, and didnt find any particular code saving or code reusability if a generic one is implemented( MUI popover is more generic already). Hence this popover is exclusively designed to be shown in the app bar(for notifications)


AlbertoPdRF:
Awesome work @san-oop! 🤩

Before diving into the code details, I have a few high level comments:
1. It seems like the pre-commit hook didn't work on your end, as there are a lot of code style issues. To make it work for future commits, on the client's directory, you have to delete the `node_modules` folder and run `npm install` again. To fix the existing issues, and also on the client's directory, you can run `npx prettier --single-quote -w *` and commit the resulting changes
   - Related to this, if you use the VS Code editor to code, the workspace configuration on the root of the repository should also help preventing these issues.
2. I think we can rename the current `Navigation` component to something like `Router`, and then rename the higher order component `NavigationBarAndDrawer` to just `Navigation`, thus matching the pattern of other sets of components throughout the code base
3. The locale files also need to be refactored to reflect the changes done on the components

After this is taken care of, we can discuss the details of the implementation! 🙂

Thanks for taking this one on! 💯

AlbertoPdRF:
Hi @san-oop!

I've seen that there's a bug: when you try to scroll down, the app crashes. The bug is related to the `HideOnScroll` function. I believe that the best way to fix this would be to merge the new `Navigation` and `NavigationBar` components into one (let's name it `NavigationBar`, this way we can rename the new `Router` component back to `Navigation`). Doing this is esentially going a step backwards in the refactoring, by just extracting the drawer and the popovers from the current `NavigationBarAndDrawer` component, but I think it will be better as things will be easier this way.

Also, please don't forget to refactor the locale files too.

Let me know if you need help with anything! :slightly_smiling_face:

AlbertoPdRF:
Hey @san-oop! Sorry for not having answered for so long. I won't maintain Tisn anymore (see #449), so I'm closing this PR.

Thank you so much for having contributed to this project, it was so nice working with you!

PR #358, opened 2020-10-25
change attandant to attendee

kailinwei:

PR #367, opened 2020-10-28
second time change, attendent to attendee; remove whitespaces and return the blankspace. 

kailinwei:


AlbertoPdRF:
Same here as on #358

